PR #11, opened 2023-09-14
Chores grooming

pavelkrolevets:


integration_test/integration_test.go
```diff
@@ -123,32 +142,37 @@ func TestHappyFlow(t *testing.T) {
...
-		srv9 := CreateServer(t, 9)
-		ops[9] = client.Operator{srv9.srv.URL, 9, &srv9.privKey.PublicKey}
-		srv10 := CreateServer(t, 10)
-		ops[10] = client.Operator{srv10.srv.URL, 10, &srv10.privKey.PublicKey}
-		srv11 := CreateServer(t, 11)
-		ops[11] = client.Operator{srv11.srv.URL, 11, &srv11.privKey.PublicKey}
-		srv12 := CreateServer(t, 12)
-		ops[12] = client.Operator{srv12.srv.URL, 12, &srv12.privKey.PublicKey}
-		clnt := client.New(ops)
+		ops := make(map[uint64]initiator.Operator)
```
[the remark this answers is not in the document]
pavelkrolevets: understood. Added negative and one positive threshold test to integration_test

pkgs/initiator/initiator.go
```diff
@@ -0,0 +1,790 @@
...
+}
+
+func (c *Initiator) NewID() [24]byte { // random ID for each new DKG initiation
+	var id [24]byte
+	b := uuid.New()
+	copy(id[:12], b[:])
+	b = uuid.New()
+	copy(id[12:], b[:])
+	return id
+}
```
y0sher, line 744: I moved this function to crypto or util, it doesn't need anything from `Initiator` so doesn't need to be a part of it.

pkgs/initiator/initiator.go
```diff
@@ -0,0 +1,790 @@
...
+			return nil, err
+		}
+		// Verify that incoming messages have valid DKG ceremony ID
+		if !bytes.Equal(id[:], tsp.Message.Identifier[:]) {
+			return nil, fmt.Errorf("incoming message has wrong ID, aborting... operator %d, msg ID %x", tsp.Signer, tsp.Message.Identifier[:])
+		}
+		// Verification operator signatures
+		if err := c.VerifyFunc(tsp.Signer, signedBytes, tsp.Signature); err != nil {
+			return nil, err
+		}
```
y0sher, line 330: I moved verification out of this function to VerifyAll, it seems  now we verify twice

pkgs/dkg/drand.go
```diff
@@ -512,3 +499,15 @@ func (o *LocalOwner) broadcastError(err error) {
...
+func (o *LocalOwner) VerifyInitiatorMessage(msg []byte, sig []byte) error {
+	pubKey, err := crypto.EncodePublicKey(o.InitiatorPublicKey)
+	if err != nil {
+		return err
+	}
+	if err := crypto.VerifyRSA(o.InitiatorPublicKey, msg, sig); err != nil {
+		return fmt.Errorf("failed to verify a message from initiator: %x", pubKey)
+	}
+	o.Logger.Info("Successfully verified initiator message signature", zap.Uint64("from", o.ID))
+	return nil
```
y0sher, line 512: `VerifyInitiatorMessage` is used only in `state` hence it doesn't make sense to have it here, not the function and not the public key itself.

pkgs/dkg/drand.go
```diff
@@ -102,44 +101,47 @@ type LocalOwner struct {
...
-	OpPrivKey  *rsa.PrivateKey
-	Owner      [20]byte
-	Nonce      uint64
+	Logger             *zap.Logger
+	ID                 uint64
+	BroadcastF         func([]byte) error
+	Suite              pairing.Suite
+	VerifyFunc         func(id uint64, msg, sig []byte) error
+	SignFunc           func([]byte) ([]byte, error)
+	OpPrivKey          *rsa.PrivateKey
```
y0sher, line 117: If we pass in SignFunc we don't need OpPrivKey, or vice verca, at first I made it a point not to pass in the key and just use the sign function.

pkgs/dkg/drand.go
```diff
@@ -102,44 +101,47 @@ type LocalOwner struct {
 	VerifyFunc func(id uint64, msg, sig []byte) error
 	SignFunc   func([]byte) ([]byte, error)
 
-	Owner common.Address
-	Nonce uint64
-	done  chan struct{}
+	Owner              common.Address
+	Nonce              uint64
+	done               chan struct{}
+	InitiatorPublicKey *rsa.PublicKey
```
y0sher, line 107: this can be part of the wrapper that we have in `state` for the local owner, since the usage is only there it doesn't make sense to put it in here.

y0sher (changes requested):
see my comments

pkgs/initiator/initiator.go
pavelkrolevets: yeah, I saw it at crypto, but got lost after merge, Ill move it to crypto again

pavelkrolevets: moved to crypto

pkgs/initiator/initiator.go
pavelkrolevets: removed

pkgs/dkg/drand.go
pavelkrolevets: I ill move it to instWrapper, as you suggested.

pkgs/dkg/drand.go
pavelkrolevets: moved to instWrapper at state

pkgs/dkg/drand.go
pavelkrolevets: we use it also to encrypt the share 
https://github.com/bloxapp/ssv-dkg/blob/chores_grooming/pkgs/dkg/drand.go#L250

pavelkrolevets: I can add EncryptFunc instead?

pavelkrolevets: Added EncryptFunc, DecryptFunc, and rsa pub instead of holding private at each instance

y0sher (approved):
Great job!
@moshe-blox @andrew-blox please take final look so we can merge

pkgs/crypto/crypto.go
```diff
@@ -442,3 +445,55 @@ func VerifyPartialSigs(sigShares map[uint64]*bls.Sign, sharePks map[uint64]*bls.
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to read file: %w", err)
+	}
+
+	operatorKeyByte, err := base64.StdEncoding.DecodeString(string(data))
+	if err != nil {
+		return nil, err
+	}
+	block, _ := pem.Decode(operatorKeyByte)
+	// TODO: resolve deprecation https://github.com/golang/go/issues/8860
```
andrew-blox, line 474: Should we resolve this?

pkgs/crypto/crypto.go
```diff
@@ -442,3 +445,55 @@ func VerifyPartialSigs(sigShares map[uint64]*bls.Sign, sharePks map[uint64]*bls.
...
+	enc := x509.IsEncryptedPEMBlock(block) //nolint
+	b := block.Bytes
+	if enc {
+		var err error
+		// TODO: resolve deprecation https://github.com/golang/go/issues/8860
```
andrew-blox, line 479: Should we resolve this?

andrew-blox (approved):

y0sher (approved):

cli/initiator/initiator.go
```diff
@@ -29,138 +30,209 @@ func init() {
...
+			logger.Fatal("😥 Please provide a valid fork name: mainnet, prater, or now_test_network")
 		}
 		owner := viper.GetString("owner")
 		if owner == "" {
-			logger.Fatal("failed to get owner address flag value", zap.Error(err))
+			logger.Fatal("😥 Failed to get owner address flag value: ", zap.Error(err))
 		}
 		nonce := viper.GetUint64("nonce")
-		withdrawPubKey, err := hex.DecodeString(withdrawAddr)
+		withdrawPubKey := common.HexToAddress(withdrawAddr).Bytes()
```
moshe-blox, line 215: This is dangerous for user input because it fails silently, so you can pass just a "0" and this would not fail.

WDYT about making our own parsing func that would return an error if it's too long, too short or not a valid hex?

pkgs/operator/operator.go
```diff
@@ -132,10 +134,14 @@ func New(key *rsa.PrivateKey) *Server {
 }
 
 func (s *Server) Start(port uint16) error {
-	s.Logger.Infof("server listening for incoming requests on port %d", port)
 	srv := &http.Server{Addr: fmt.Sprintf(":%v", port), Handler: s.Router}
```
moshe-blox, line 137: Not part of this PR, but it'd be nice if we can add timeouts to the `http.Server`

pkgs/operator/state.go
```diff
@@ -0,0 +1,280 @@
...
+		PrivateKey:       pv,
+	}
+}
+
+func (s *Switch) InitInstance(reqID [24]byte, initMsg *wire.Transport, initiatorSignature []byte) ([]byte, error) {
+	logger := s.Logger.With(zap.String("reqid", hex.EncodeToString(reqID[:])))
+	logger.Info("🚀 Initializing DKG instance")
+	init := &wire.Init{}
+	if err := init.UnmarshalSSZ(initMsg.Data); err != nil {
+		return nil, err
```
moshe-blox, line 173: We should return error wrappings (as you did below in `VerifyRSA`) here and in the rest of the code, so we can tell where exactly things failed in retrospect.

moshe-blox (changes requested):
Looks good 👍

Left a few comments, although did not finish reviewing yet.

pkgs/operator/operator.go
pavelkrolevets: We have Chi limiting requests to routes, to prevent DDOS, if I understood you correctly

cli/initiator/initiator.go
pavelkrolevets: added HexToAddress with error handling to utils + unit tests for it

pkgs/operator/state.go
pavelkrolevets: added error wrapps at state. Should we have wrapps everywhere?

y0sher (approved):

lior-blox:
Merged as part 1 of grooming

PR #16, opened 2023-09-29
Key resharing

pavelkrolevets:
- resharing flow
- badger DB to store DKG results


pavelkrolevets:
@lior-blox @y0sher 

We have some sort of a problem.

If we chose a set of new operators which are disjoint from old set, we get two distinct groups of operators able to sign for same validator public key. Basically, when we reshare, we create new shares with the same public key point, so both sets of operators can create a signature valid for this point.

In the same time. Operators from different disjoint sets cannot me mixed to create a valid master signature. Only a subset more or equal Threshold of each disjoint set can create a valid master signature. 

Considering this, we can warn a user about this fact. IMHO a solution can be to reshare to a joint set of new operators.

y0sher:
Please rebase this on main, I'm reviewing meanwhile.

README.md
```diff
@@ -313,6 +324,52 @@ ssv-dkg init --configPath ./initiator-config/initiator.yaml
...
+          --oldID "dbd12b3155454666a6710a2262695bb82cda41948d612d98" \
+          --operatorsInfoPath ./examples/operators_integration.json \
+          # Alternatively:
+          # --operatorsInfo: '[{"id": 1,"public_key": "LS0tLS1CRUdJTiBSU0....","ip": "http://localhost:3030"}, {"id": 2,"public_key": "LS0tLS1CRUdJTiBSU0....","ip": "http://localhost:3030"},...]'
+          --owner 0x81592c3de184a3e2c0dcb5a261bc107bfa91f494 \
+          --nonce 4 \
+          --network "mainnet" \
+          --outputPath /output \
+          --initiatorPrivKey ./encrypted_private_key.json \
+          --initiatorPrivKeyPassword ./password \
```
y0sher, line 349: we should probably add a note about that generating a new key here isn't possible since we want to use the previous key used to init.

cli/initiator/reshare.go
```diff
@@ -0,0 +1,266 @@
+package initiator
```
y0sher, line 1: Its seems like this file shares a lot of code with `init.go`, can resharing be a subcommand of init? or at least lets break out some of the repeated code into functions, the purpose is that we won't have to update the same code twice (or forget to) if we do any changes.

pkgs/dkg/drand.go
```diff
@@ -159,27 +196,181 @@ func (o *LocalOwner) StartDKG() error {
...
+		NewT:         int(o.Data.Reshare.NewT),
+		Board:        o.Board,
+		PublicCoeffs: coefs,
+		Logger:       logger,
+	})
+	if err != nil {
+		return err
+	}
+	for _, b := range o.Deals {
+		o.Board.DealC <- *b
```
y0sher, line 356: I don't get how this works honestly as from what I se in the code of `board.go` this channel is unbuffered. I guess starting the protocol reads these from `p`, but it means it might be smarter to run the go routine underneath before pushing into the channel.

y0sher: well I guess `NewReshareProtocolNewNodes` listen to this so no need to run the goroutine before.

pkgs/initiator/initiator.go
```diff
@@ -751,3 +949,22 @@ func (c *Initiator) GetThreshold(ids []uint64) (int, error) {
 	threshold := len(ids) - ((len(ids) - 1) / 3)
 	return threshold, nil
 }
+
+func (c *Initiator) JoinSets(oldOperators []*wire.Operator, newOperators []*wire.Operator) []*wire.Operator {
```
y0sher, line 953: no reason that this func will be a member of `Initiator`. it can be stand alone function and more testable like that.

pkgs/operator/state.go
```diff
@@ -220,11 +309,139 @@ func (s *Switch) InitInstance(reqID [24]byte, initMsg *wire.Transport, initiator
...
+	inst, resp, err := s.CreateInstanceReshare(reqID, reshare, initiatorPubKey, nil)
+	if err != nil {
+		return nil, err
+	}
+	s.Mtx.Lock()
+	_, ok = s.Instances[reqID]
+	if ok {
+		s.Mtx.Unlock()
+		return nil, ErrAlreadyExists
+	}
```
y0sher, line 437: there's a bug here. in this case the instance will be created and run but not returned. at least return the exiting one.

pkgs/operator/state.go
```diff
@@ -220,11 +309,139 @@ func (s *Switch) InitInstance(reqID [24]byte, initMsg *wire.Transport, initiator
...
+
+	s.Logger.Info("Starting resharing protocol")
+	// try to get old share local owner first
+	var shareFromDB basedb.Obj
+	secret := &kyber_dkg.DistKeyShare{}
+	shareFromDB, ok, err := s.DB.Get([]byte("secret"), reshare.OldID[:])
+	if err != nil {
+		return nil, err
+	}
+	if ok {
```
y0sher, line 346: we aren't going to reshare if there's no secret in db right? lets do 
```golang
if !ok {
 fmt.Errorf("can't find secret share in db for id %v", reshare.OldID[:]) 
}
```

README.md
pavelkrolevets: Agree. Ill add a note to readme.

cli/initiator/reshare.go
pavelkrolevets: Honestly, I would rather keep it separate commands. I tried to have resharing as a subcommand, but it looks messy together. Ill separate functions as much as I can find

pkgs/operator/state.go
pavelkrolevets: this part is redundant, removed

pkgs/operator/state.go
pavelkrolevets: Not exactly, new operators may not have an old share in DB so we are creating an instance without a share

cli/initiator/reshare.go
```diff
@@ -0,0 +1,231 @@
...
+
+	"github.com/bloxapp/ssv-dkg/cli/flags"
+	cli_utils "github.com/bloxapp/ssv-dkg/cli/utils"
+	"github.com/bloxapp/ssv-dkg/pkgs/crypto"
+	"github.com/bloxapp/ssv-dkg/pkgs/initiator"
+	"github.com/bloxapp/ssv-dkg/pkgs/utils"
+)
+
+func init() {
+	flags.InitiatorPrivateKeyFlag(StartReshare)
```
y0sher, line 22: for example all this code is very repeatable, all this functions are the same just with different const, maybe we should also just pass a param in instead of having a function for each?

cli/initiator/reshare.go
```diff
@@ -0,0 +1,231 @@
...
+	flags.NewOperatorIDsFlag(StartReshare)
+	flags.OwnerAddressFlag(StartReshare)
+	flags.NonceFlag(StartReshare)
+	flags.ConfigPathFlag(StartReshare)
+	flags.LogLevelFlag(StartReshare)
+	flags.LogFormatFlag(StartReshare)
+	flags.LogLevelFormatFlag(StartReshare)
+	flags.LogFilePathFlag(StartReshare)
+	flags.ResultPathFlag(StartReshare)
+	if err := viper.BindPFlag("withdrawAddress", StartReshare.PersistentFlags().Lookup("withdrawAddress")); err != nil {
```
y0sher, line 37: also here I think it's best to move this to a function as since its repeated a lot. also we're using the raw strings for the flag names and they are defined as consts in `flags`, better use the consts so we don't have to change many times if something has changed.

y0sher: honestly this is more suitable to happen in #20 , but these PRs and kinda overlapping now

cli/initiator/reshare.go
```diff
@@ -0,0 +1,231 @@
...
+			if err != nil {
+				logger.Fatal(err.Error())
+			}
+		}
+		if err := viper.BindPFlag("operatorIDs", cmd.Flags().Lookup("operatorIDs")); err != nil {
+			logger.Fatal("😥 Failed to bind a flag: ", zap.Error(err))
+		}
+		oldParticipants := viper.GetStringSlice("operatorIDs")
+		if oldParticipants == nil {
+			logger.Fatal("😥 Failed to get operator IDs flag value: ", zap.Error(err))
```
y0sher, line 152: most of this code is repeated, for example loading the output path and verifying it is the same code as `initator/init.go`, should be take out to a separate function (probably in the `initiator` package) and used in both,
same goes for operators info/info path.
participants code is just a tad different because of the new and old participants here, but it already has loadParticipants for that.

cli/initiator/reshare.go
```diff
@@ -0,0 +1,231 @@
...
+		keyShares, err := dkgInitiator.StartReshare(id, oldID, oldParts, newParts, common.HexToAddress(owner), nonce)
+		if err != nil {
+			logger.Fatal("😥 Failed to initiate DKG ceremony: ", zap.Error(err))
+		}
+		// Save results
+		keysharesFinalPath := fmt.Sprintf("%s/keyshares-reshared-%v-%v.json", outputPath, keyShares.Payload.PublicKey, hex.EncodeToString(id[:]))
+		logger.Info("💾 Writing keyshares payload to file", zap.String("path", keysharesFinalPath))
+		err = utils.WriteJSON(keysharesFinalPath, keyShares)
+		if err != nil {
+			logger.Warn("Failed writing keyshares file: ", zap.Error(err))
```
y0sher, line 210: - repeated code
- filename should probably use a const

pkgs/dkg/drand.go
```diff
@@ -81,29 +93,53 @@ func (msg *Result) Decode(data []byte) error {
...
-	startedDKG  chan struct{}
+	StartedDKG  chan struct{}
 	ErrorChan   chan error
 	ID          uint64
-	data        *DKGData
-	b           *board.Board
-	suite       pairing.Suite
+	Data        *DKGData
+	Board       *board.Board
+	Suite       pairing.Suite
```
y0sher, line 126: why all of these are public now?, not a big deal but I'm not sure there's a good reason.

pkgs/dkg/drand.go
```diff
@@ -159,27 +203,181 @@ func (o *LocalOwner) StartDKG() error {
...
+		}
+
+		OldNodes = append(OldNodes, dkg.Node{
+			Index:  dkg.Index(op.ID - 1),
+			Public: p,
+		})
+	}
+	o.Logger.Debug("Staring DKG with nodes: ")
+	for _, n := range append(OldNodes, NewNodes...) {
+		o.Logger.Debug("node: ", zap.String("nodes", n.Public.String()))
```
y0sher, line 262: is the field here the public key? why is it called `nodes`?

pkgs/dkg/drand.go
```diff
@@ -159,27 +203,181 @@ func (o *LocalOwner) StartDKG() error {
...
+		p := o.Suite.G1().Point()
+		if err := p.UnmarshalBinary(e.PK); err != nil {
+			return err
+		}
+
+		NewNodes = append(NewNodes, dkg.Node{
+			Index:  dkg.Index(op.ID - 1),
+			Public: p,
+		})
+	}
```
y0sher, line 243: I've seen this same code block at least 3 times, worth having a dedicated function that recvs the operators map `newOperators` in this case and returns the nodes, `newNodes` in this case)

pkgs/dkg/drand.go
```diff
@@ -159,27 +203,181 @@ func (o *LocalOwner) StartDKG() error {
...
+		OldNodes = append(OldNodes, dkg.Node{
+			Index:  dkg.Index(op.ID - 1),
+			Public: p,
+		})
+	}
+	o.Logger.Debug("Staring resharing DKG with nodes: ")
+	for _, n := range append(OldNodes, NewNodes...) {
+		o.Logger.Debug("node: ", zap.String("nodes", n.Public.String()))
+	}
+	coefsBytes := utils.SplitBytes(commits, 48)
```
y0sher, line 335: coefs should also be defined here since its only used now.

pkgs/dkg/drand.go
```diff
@@ -159,27 +203,181 @@ func (o *LocalOwner) StartDKG() error {
...
+		return err
+	}
+	for _, b := range o.Deals {
+		o.Board.DealC <- *b
+	}
+	go func(p *dkg.Protocol, postF func(res *dkg.OptionResult) error) {
+		res := <-p.WaitEnd()
+		postF(&res)
+	}(p, o.PostReshare)
+	// close(o.startedDKG)
```
y0sher, line 369: if this is not needed remove it.

pkgs/dkg/drand.go
```diff
@@ -431,34 +799,117 @@ func (o *LocalOwner) Process(from uint64, st *wire.SignedTransport) error {
...
+			return ErrAlreadyExists
+		}
+
+		o.Exchanges[from] = exchMsg
+		allOps := append(o.Data.Reshare.OldOperators, o.Data.Reshare.NewOperators...)
+		for _, op := range allOps {
+			if o.Exchanges[op.ID] == nil {
+				return nil
+			}
+		}
```
y0sher, line 823: better approach like in normal dkg is just to check whether the len is equal.
`len(o.Exchanges) == allOps`, more efficient even though the map is small.

pkgs/operator/state.go
```diff
@@ -220,11 +309,78 @@ func (s *Switch) InitInstance(reqID [24]byte, initMsg *wire.Transport, initiator
...
+	}
+	_, ok = s.Instances[reqID]
+	if ok {
+		tm := s.InstanceInitTime[reqID]
+		if !time.Now().After(tm.Add(MaxInstanceTime)) {
+			s.Mtx.Unlock()
+			return nil, ErrAlreadyExists
+		}
+		delete(s.Instances, reqID)
+		delete(s.InstanceInitTime, reqID)
```
y0sher, line 369: this is potential attack vector I just thought of ( though it was probably my bad ). if a reqid is expired like in this code and someone runs a new instance with this reqid it might rewrite the secret share? are we saving them by id? we should think how to mitigate, might be to not allow reusing a reqid, if its in the list or if its in the db.

pkgs/operator/utils.go
y0sher, line 1: I would rather the test functions be in separate file/package.

pkgs/wire/dkg.go
```diff
@@ -71,6 +110,36 @@ func NewDKGProtocol(config *Config) (*dkg.Protocol, error) {
 
 	return ret, nil
 }
+func NewReshareProtocolNewNodes(config *Config) (*dkg.Protocol, error) {
+	dkgLogger := New(config.Logger)
+	dkgConfig := &dkg.Config{
+		Longterm:     config.Secret,
```
y0sher, line 116: how does the new nodes have secret?

y0sher (changes requested):
good work on this one figuring out all the quirks, see my comments.

pkgs/wire/dkg.go
pavelkrolevets: This id not a secret share. This a secret scalar used in DKG for encrypting/decrypting deals

pkgs/dkg/drand.go
pavelkrolevets: I was trying to use it at another modules. Just left it exported eventually

pkgs/operator/state.go
pavelkrolevets: Yeah looks like its a weak point here. We do store a secret share at DB https://github.com/bloxapp/ssv-dkg/blob/key_resharing/pkgs/dkg/drand.go#L442 using `reqID` as we need it at resharing. 
An dishonest initiator seems can rewrite a share at operators if a new init ceremony with the same ID is being run, but the resulting validator will be different anyway. So looks like the only resharing at risk here.

pavelkrolevets: I will add the check if we already have ID at DB - we bail right away.

pkgs/dkg/drand.go
pavelkrolevets: wanted to use it, but not. Left exported.

pkgs/dkg/drand.go
```diff
@@ -22,33 +18,46 @@ import (
...
 const (
 	// MaxEffectiveBalanceInGwei is the max effective balance
 	MaxEffectiveBalanceInGwei phase0.Gwei = 32000000000
-
 	// BLSWithdrawalPrefixByte is the BLS withdrawal prefix
 	BLSWithdrawalPrefixByte = byte(0)
 )
 
 // IsSupportedDepositNetwork returns true if the given network is supported
 var IsSupportedDepositNetwork = func(network eth2_key_manager_core.Network) bool {
```
y0sher, line 40: I'm pretty sure I removed this in my holesky_support, this function exist in `eth2_key_manager` so its better to use it from there than rewriting.

pavelkrolevets: I merged to here `holesky_support`. This is probably a leftover. Ill remove

pkgs/dkg/drand.go
```diff
@@ -22,33 +15,41 @@ import (
...
+	"github.com/bloxapp/ssv-dkg/pkgs/crypto"
+	"github.com/bloxapp/ssv-dkg/pkgs/utils"
+	"github.com/bloxapp/ssv-dkg/pkgs/wire"
+	ssvspec_types "github.com/bloxapp/ssv-spec/types"
+	"github.com/bloxapp/ssv/storage/kv"
 )
 
+var OutputPath string
+var StoreShare bool
+
```
y0sher, line 30: why are these here? why aren't we getting them from the config like all the rests of the parameters?

pkgs/dkg/drand.go
```diff
@@ -214,77 +360,199 @@ func (o *LocalOwner) PostDKG(res *dkg.OptionResult) error {
...
-			Secret: secretKeyBLS.SerializeToHexStr(),
+	if StoreShare {
+		ciphertext, err := o.encryptSecretShare(secretKeyBLS)
+		if err != nil {
+			o.Logger.Error("Cant encrypt secret share: ", zap.Error(err))
+			o.broadcastError(err)
+			return err
 		}
-		err = utils.WriteJSON("./secret_share_"+hex.EncodeToString(o.data.ReqID[:]), &data)
+		err = utils.StoreSecretShareToFile(OutputPath, res.Result.Key.Share.I, ciphertext, validatorPubKey)
```
y0sher, line 414: same as I said about the db, I don't like the fact we're writing files from here. so far we returned from here and used higher level code to store stuff. I want to keep this code as pure protocol and not do any I/O.

pkgs/dkg/drand.go
```diff
@@ -214,77 +360,199 @@ func (o *LocalOwner) PostDKG(res *dkg.OptionResult) error {
...
-	if !val {
+	// Get BLS partial secret key share from DKG
+	secretKeyBLS, err := crypto.ResultToShareSecretKey(res.Result)
+	if err != nil {
 		o.broadcastError(err)
-		return fmt.Errorf("partial deposit root signature is not valid %x", depositRootSig.Serialize())
+		return err
+	}
+	// Store secret if requested
+	if StoreShare {
```
y0sher, line 535: same comment as above

pkgs/dkg/drand.go
y0sher: Anything exported should be commented about how it is used. if it is not used and not meant to be used. unexport it. if it is meant to be used add a comment about how it is used. this is about everything exported in this package.

pkgs/dkg/drand.go
pavelkrolevets: Removed.

pkgs/dkg/drand.go
```diff
@@ -214,77 +360,199 @@ func (o *LocalOwner) PostDKG(res *dkg.OptionResult) error {
...
+	if err != nil {
+		o.broadcastError(err)
+		return err
+	}
+	encBin, err := o.EncryptSecretDB(bin)
+	if err != nil {
+		o.broadcastError(err)
+		return err
+	}
+	err = o.DB.Set([]byte("secret"), o.Data.ReqID[:], encBin)
```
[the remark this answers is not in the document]
pavelkrolevets: Moved higher level to operator/state package.

pkgs/dkg/drand.go
pavelkrolevets: Moved higher level to operator/state package.

pkgs/dkg/drand.go
pavelkrolevets: Moved higher level to operator/state package.

y0sher (approved):

PR #17, opened 2023-10-12
update initiator flow

pavelkrolevets:
- two flags for operator data: either path to operators info file, or path to the folder where to look for `operators_info.json`
- either provide RSA key file, or generate RSA and store at output directory


cli/flags/flags.go
```diff
@@ -12,11 +12,13 @@ const (
 	threshold                = "threshold"
 	withdrawAddress          = "withdrawAddress"
 	operatorIDs              = "operatorIDs"
-	operatorsInfo            = "operatorsInfoPath"
+	operatorsInfoFile        = "operatorsInfoFile"
```
y0sher, line 15: this name is a bit misleading, lets make it `operatorsInfo`, make sure to add in the help section we're expecting a json format string.

cli/initiator/initiator.go
```diff
@@ -237,6 +320,19 @@ var StartDKG = &cobra.Command{
 		if err != nil {
 			logger.Warn("Failed writing keyshares file: ", zap.Error(err))
 		}
+		if privKeyPath == "" && generateInitiatorKey {
+			rsaKeyPath := fmt.Sprintf("%s/encrypted_private_key-%v.json", keysharesOutputPath, depositData.PubKey)
```
y0sher, line 324: that's a good way to use the same output path, maybe we should also combine the deposit and keyshares path args to just one arg `outputPath`

y0sher (changes requested):
please see my comments, mainly the misunderstading about the operators info.

cli/initiator/initiator.go
pavelkrolevets: Yeah, I was thinking about that as well. Will make only one output path.

cli/initiator/initiator.go
```diff
@@ -119,33 +133,66 @@ var StartDKG = &cobra.Command{
...
+				filePath := operatorsInfoPath + "operators_info.json"
+				if _, err := os.Stat(filePath); os.IsNotExist(err) {
+					logger.Fatal("😥 Failed to find operator info file at provided path: ", zap.Error(err))
+				}
+				opsfile, err := os.ReadFile(filePath)
+				if err != nil {
+					logger.Fatal("😥 Failed to read operator info file:", zap.Error(err))
+				}
+				opMap, err = initiator.LoadOperatorsJson(opsfile)
+				if err != nil {
```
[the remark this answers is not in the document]
pavelkrolevets: Yeah, seems misunderstood. Will make a string incoming

cli/initiator/initiator.go
```diff
@@ -106,46 +114,71 @@ var StartDKG = &cobra.Command{
...
-			logger.Fatal("😥 Failed to load operators: ", zap.Error(err))
+		var opMap initiator.Operators
+		if operatorsInfo != "" {
+			logger.Info("📖 reading raw JSON string of operators info", zap.String("", operatorsInfo))
+			opMap, err = initiator.LoadOperatorsJson([]byte(operatorsInfo))
+			if err != nil {
+				logger.Fatal("😥 Failed to load operators: ", zap.Error(err))
+			}
+		}
+		if operatorsInfoPath != "" {
```
y0sher, line 152: what if both of them are empty?

examples/initiator/operators_data/operators_info.json
```diff
@@ -0,0 +1,22 @@
+[
```
y0sher, line 1: are we adding another file? if this is used now instead of another lets remove the previous.

y0sher (changes requested):
last small comments

cli/initiator/initiator.go
pavelkrolevets: We have a check for it: https://github.com/bloxapp/ssv-dkg/blob/after_demo_fixes/cli/initiator/initiator.go#L138

y0sher (approved):

pkgs/dkg/drand_test.go
```diff
@@ -0,0 +1,228 @@
...
+
+func (ts *testState) ForOld(f func(o *LocalOwner) error, oldOps []*wire2.Operator) error {
+	for _, op := range oldOps {
+		newOp := ts.ops[op.ID]
+		if err := f(newOp); err != nil {
+			return err
+		}
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
y0sher: and they are not used

y0sher (approved):

y0sher (changes requested):
- Please make sure its possible to use all cli args and not load a config (like the example)
- Make sure the `operatorsInfo` flag works and update the correct actual format in the readme
- Lets reduce readme changes to only changes required and related to the changes we do. for example breaking the `design.md` out of the code should not be a part of this PR.

PR #29, opened 2023-11-24
enable native bulk ceremony capability

pavelkrolevets:
* added a reference of the operator ID in the dial tcp timeout error

* renamed output files to:
```
- deposit-[validator_pubkey] to deposit_data-[validator_pubkey]
- encrypted_private_key-[validator_pubkey].json to initiator_encrypted_key.json
- password-[validator_pubkey] to initiator_password
- keyshares-[validator_pubkey] to keyshares-[validator_pubkey]-[ownerNonce]-[instance-id]
```
* use `configPath` instead of `initiatorPrivKey` &` initiatorPrivKeyPassword` (both in initiator.yaml and as arguments)

 * `--generateInitiatorKey` renamed to `--generateInitiatorKeyIfNotExisting`
The argument updated to generate a new initiator key and password only if they are not existing.
Exists - use them for the ceremony
Does not exist - generate new
The argument output the generated key and password to a folder named config
```
`configPath` folder
├── initiator_encrypted_key.json
└── initiator_password
```
* added `instance_id.json` that include the instance_id which was used for the ceremony

* added `bulk key generation`  feature to initiator to create up to 100 validators in parallel. Use flag:

```
--validators [1-100]
```

* ceremony files generated within a folder named ceremony-[timestamp]
```
ceremony-[timestamp] folder
├── [validator-pubkey] x 100 folders
 |         ├── deposit_data-[validator-pubkey]json
 |         ├── keyshares-[validator-pubkey]-[owner-nonce]-[instance-id].json
 |         └── instance_id.json
├── deposit_data.json # aggregated in case of more than 1 validator generated
├── keyshares.json  # aggregated in case of more than 1 validator generated
└── instance_id.json  # aggregated in case of more than 1 validator generated
```

* added  generated deposit data and keyshares file output to all ceremony participants (all operators)

```
outputs at operators generated within a folder named ceremony-[timestamp]:
ceremony-[timestamp] folder
├── deposit_data-[validator_pubkey].json
├── keyshares-[validator_pubkey]-[owner-nonce]-[instance-id].json
└── instance-id.json
```

* added a new live check command 
```sh
ssv-dkg ping \
          --operatorIDs 1,2,3,4 \
          --operatorsInfoPath ./examples/operators_integration.json \
          --configPath ./examples/config
```
* added initiator signature of ceremony ID that operators can reshare only with the same initiator who `init`

* added version integrity check for initiator and operators


pavelkrolevets:
TODO: mutex

pkgs/operator/state.go
```diff
@@ -183,12 +185,24 @@ func (s *Switch) CreateInstanceReshare(reqID [24]byte, reshare *wire.Reshare, in
...
+			}
+			secretShare, err = s.GetSecretShare(reshare.OldID, pub)
+			if err != nil {
+				return nil, nil, err
+			}
+			owner.SecretShare = secretShare
+			for _, point := range secretShare.Commits {
+				b, _ := point.MarshalBinary()
+				commits = append(commits, b...)
+			}
```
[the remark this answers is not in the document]
pavelkrolevets: We can have only one operator ID in the list of old operators. But yes we can break the iter if found operator ID in the list

pkgs/operator/state.go
```diff
@@ -531,3 +517,160 @@ func (s *Switch) GetSecretShare(id [24]byte) (*kyber_dkg.DistKeyShare, error) {
...
+	marshalledWireMsg, err := incMsg.MarshalSSZ()
+	if err != nil {
+		return nil, err
+	}
+	err = crypto.VerifyRSA(initiatorPubKey, marshalledWireMsg, incMsg.Signature)
+	if err != nil {
+		return nil, fmt.Errorf("init message: initiator signature isn't valid: %s", err.Error())
+	}
+	operatorPubKey := s.PrivateKey.Public().(*rsa.PublicKey)
+	pkBytes, err := crypto.EncodePublicKey(operatorPubKey)
```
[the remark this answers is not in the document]
pavelkrolevets: done. Agree this should improve as we dont need to work on it each time.

cli/flags/flags.go
```diff
@@ -136,14 +137,14 @@ func PrivateKeyFlag(c *cobra.Command) {
...
-// GenerateInitiatorKeyFlag adds flag to generate a random secure password and initiator RSA key pair encrypted with this password
-func GenerateInitiatorKeyFlag(c *cobra.Command) {
-	AddPersistentBoolFlag(c, generateInitiatorKey, false, "Generates a random secure password and initiator RSA key pair encrypted with this password", false)
+// GenerateInitiatorKeyIfNotExistingFlag adds flag to generate a random secure password and initiator RSA key pair encrypted with this password
+func GenerateInitiatorKeyIfNotExistingFlag(c *cobra.Command) {
+	AddPersistentBoolFlag(c, generateInitiatorKeyIfNotExisting, false, "Generates a random secure password and initiator RSA key pair encrypted with this password", false)
 }
 
 // GetGenerateInitiatorKeyFlagValue gets flag to generate a random secure password and initiator RSA key pair encrypted with this password
 func GetGenerateInitiatorKeyFlagValue(c *cobra.Command) (bool, error) {
```
y0sher, line 146: just to say consistent this should be `GetGenerateInitiatorKeyIfNotExistingFlagValue` right?

pavelkrolevets: removed unused `Get`s

cli/initiator/initiator.go
```diff
@@ -50,56 +48,56 @@ var StartDKG = &cobra.Command{
...
+		var depositDataArr []*initiator.DepositDataJson
+		var keySharesArr []*initiator.KeyShares
+		var ids [][24]byte
+		var nonces []uint64
+		for res := range resultChan {
+			if res.err != nil {
+				logger.Fatal("😥 Failed to initiate DKG ceremony: ", zap.Error(res.err))
 			}
-			logger.Info("Private key encrypted and stored at", zap.String("path", cli_utils.OutputPath))
+			logger.Info("🎯 Received result.")
```
y0sher, line 92: I don't think we should log this for completed ceremony

cli/utils/utils.go
```diff
@@ -363,6 +373,11 @@ func BindInitFlags(cmd *cobra.Command) error {
 	if Network == "" {
 		return fmt.Errorf("😥 Failed to get fork version flag value")
 	}
+	Validators = viper.GetUint64("validators")
```
y0sher, line 376: shound't we set default to 1 if its not set (0?)

pkgs/crypto/crypto.go
```diff
@@ -593,3 +594,19 @@ func VerifyReconstructedSignature(sig *bls.Sign, validatorPubKey, msg []byte) er
 	}
 	return nil
 }
+
+func ReadEncryptedRSAKey(privKeyPath, privKeyPassPath string) (*rsa.PrivateKey, error) {
+	keyStorePassword, err := os.ReadFile(privKeyPassPath)
```
y0sher, line 599: imo we should not ready files in `crypto.go` read the file outside and pass the function the results. easier to test and use.

pkgs/dkg/drand.go
```diff
@@ -637,6 +648,9 @@ func (o *LocalOwner) processDKG(from uint64, msg *wire.Transport) error {
 
 // Process processes incoming messages from initiator at /dkg route
 func (o *LocalOwner) Process(from uint64, st *wire.SignedTransport) error {
+	if !bytes.Equal(st.Message.Version, o.version) {
```
y0sher, line 651: see comment above about the version

pkgs/initiator/initiator.go
```diff
@@ -563,11 +569,30 @@ func (c *Initiator) StartDKG(id [24]byte, withdraw []byte, ids []uint64, network
...
+	if err != nil {
+		return nil, nil, err
+	}
+	resultMsg := &wire.ResultData{
+		Operators:          ops,
+		InitiatorPublicKey: pkBytes,
+		DepositData:        depositData,
+		KeysharesData:      keysharesData,
+	}
+	err = c.sendResult(resultMsg, ops, consts.API_RESULTS_URL, id)
```
y0sher, line 592: I don't understand why do we need to send this data back to the operators. they should be able to reproduce this whole thing using info the got at init. this looks like a possible attack vector.

y0sher: we also don't want the operators to be able to use these by themselves. this belong to the initiator only. (the complete deposit and keyshare files)

pkgs/initiator/initiator.go
```diff
@@ -1015,3 +1001,91 @@ func VerifySharesData(ops map[uint64]Operator, keys []*rsa.PrivateKey, ks *KeySh
...
+	ops, err := ValidatedOperatorData(ids, c.Operators)
+	if err != nil {
+		return nil, err
+	}
+	// Add messages verification coming form operators
+	verify, err := c.CreateVerifyFunc(ops)
+	if err != nil {
+		return nil, err
+	}
+	c.VerifyFunc = verify
```
y0sher, line 1015: this is changed on other places? if yes should probably happen once on constructor.

pkgs/operator/operator.go
```diff
@@ -139,47 +132,91 @@ func RegisterRoutes(s *Server) {
...
 			}
 			logger.Info("✅ Instance started successfully")
 
 			writer.WriteHeader(http.StatusOK)
 			writer.Write(b)
 		})
+		s.Router.Route("/health_check", func(r chi.Router) {
+			r.Post("/", func(writer http.ResponseWriter, request *http.Request) {
+				rawdata, _ := io.ReadAll(request.Body)
+				signedPintMsg := &wire.SignedTransport{}
```
y0sher, line 160: type: `signedPintMsg` -> `signedPingMsg`

pkgs/operator/operator.go
```diff
@@ -139,47 +132,91 @@ func RegisterRoutes(s *Server) {
...
+				utils.WriteErrorResponse(s.Logger, writer, err, http.StatusBadRequest)
 				return
 			}
 			logger.Info("✅ Instance started successfully")
 
 			writer.WriteHeader(http.StatusOK)
 			writer.Write(b)
 		})
+		s.Router.Route("/health_check", func(r chi.Router) {
+			r.Post("/", func(writer http.ResponseWriter, request *http.Request) {
```
y0sher, line 158: sounds to me like a `GET` route and not `POST`

y0sher: well since data is being sent in ping, maybe not.

pkgs/operator/operator.go
```diff
@@ -139,47 +132,91 @@ func RegisterRoutes(s *Server) {
...
+				b, err := s.State.Pong(signedPintMsg)
+				if err != nil {
+					utils.WriteErrorResponse(s.Logger, writer, err, http.StatusBadRequest)
+					return
+				}
+				writer.WriteHeader(http.StatusOK)
+				writer.Write(b)
+			})
+		})
+		s.Router.Route("/results", func(r chi.Router) {
```
y0sher, line 180: dont like this route, operator should be the one providing the results to initiator. all operators should be able to make up the results and sign independently.

pkgs/operator/state.go
```diff
@@ -531,3 +509,155 @@ func (s *Switch) GetSecretShare(id [24]byte) (*kyber_dkg.DistKeyShare, error) {
...
+		if err != nil {
+			return 0, err
+		}
+		ops = resData.Operators
+	}
+	err = s.VerifySig(incMsg, initiatorPubKey)
+	if err != nil {
+		return 0, err
+	}
+	operatorID, err := GetOperatorID(ops, s.PubKeyBytes)
```
y0sher, line 602: I don't see a point in verifying signature of message only against a public key included in the message itself. are you sure this is what you meant to do here?

pkgs/operator/operator.go
pavelkrolevets: each operator has only his encrypted share, all enc shares combined at initiator to one keyshares file. We would need a round of exchanges between operators, or a simple message from initiator

pkgs/operator/state.go
pavelkrolevets: We follow this logic everywhere, I just didnt want to deviate

pkgs/initiator/initiator.go
pavelkrolevets: https://github.com/bloxapp/ssv-dkg/pull/29#discussion_r1415948572

pkgs/dkg/drand.go
```diff
@@ -591,14 +599,17 @@ func (o *LocalOwner) InitReshare(reqID [24]byte, reshare *wire.Reshare, commits
...
 }
 
 // processDKG after receiving a kyber message type at /dkg route
 // KyberDealBundleMessageType - message that contains all the deals and the public polynomial from participating party
 // KyberResponseBundleMessageType - status for the deals received at deal bundle
 // KyberJustificationBundleMessageType - all justifications for each complaint for received deals bundles
 func (o *LocalOwner) processDKG(from uint64, msg *wire.Transport) error {
+	if !bytes.Equal(msg.Version, o.version) {
+		return utils.ErrVersion
+	}
```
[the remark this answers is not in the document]
pavelkrolevets: I nested version check deep that all messages have consistent version.

pkgs/operator/operator.go
pavelkrolevets: same with deposit data, we have a master sig which is a combination of partial sigs. No operator has all partial sigs at the end of the initial ceremony.

pkgs/operator/operator.go
pavelkrolevets: done

pkgs/initiator/initiator.go
pavelkrolevets: done

cli/initiator/initiator.go
pavelkrolevets: removed

pkgs/dkg/drand.go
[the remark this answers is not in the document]
y0sher: do we suspect someone is going to a version in the middle of the dkg process? IMO its enough to check on the "handshake" which for us is the init.  but we can consult with @MatusKysel @lior-blox @moshe-blox

pkgs/operator/operator.go
y0sher: why do they need them though? this data is needed for the initiator to deposit and register the validator to ssv. the operators don't need it. or else I'm missing something. 
@lior-blox / @MatusKysel  - 
do we want operators to have the complete data? isn't that a security concern? shouldn't it belong only to the initiator?

pkgs/operator/state.go
y0sher: If we do this everywhere then this is really bad. on the operators side we should validate the signature using the initiator pubkey we received on init and not in every message, this doesn't add any security like this.
also on the initiator (or on operators when they extract every operator message from the combined message) side we should validate the messages using the pubkey of an operator that claims to have signed the message. we should have them from beforehand in the INIT . in fact its not a must to have pubkey inside the message, an ID will be enough and then we verify we the pubkey we know belongs to that ID.

pavelkrolevets: - we can move initiator pub to instance at operator side, then we will have a relation between instance ID and initiator`s pub key
- at initiator side we have operator pubs so able to verify operator message sigs

Yes, we can get rid of pub keys at messages.

pavelkrolevets: Update: 
in case of Result messages we have already verification of ID + pub. Ill reuse it
In case of ping/pong for health check ill keep it, as we dont have instance related to it.

pkgs/dkg/drand.go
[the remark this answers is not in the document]
pavelkrolevets: okay, ill leave only one check

pkgs/operator/state.go
y0sher: I still think validating the key that is inside the message without comparing it to anything doesn't add any security, anyone can create a valid key and sign with it.

pavelkrolevets: Okay, then do we need any security doing health_check? Probably no for initiator side. I changed to GET requests and verification of operator sigs

cli/initiator/initiator.go
```diff
@@ -30,76 +28,76 @@ var StartDKG = &cobra.Command{
...
-				err = os.WriteFile(rsaKeyPasswordPath, []byte(password), 0o644)
-				if err != nil {
-					logger.Fatal("Failed to write encrypted private key to file", zap.Error(err))
+		var wg sync.WaitGroup
+		resultChan := make(chan *Result)
+		for i := 0; i < int(cli_utils.Validators); i++ {
+			wg.Add(1)
+			go func(i int) {
+				defer wg.Done()
+				// create a new ID
```
y0sher, line 68: IMO you could achieve the same thing with simpler concurrency patterns and only a channel.

cli/utils/utils.go
```diff
@@ -308,6 +307,13 @@ func BindInitiatorBaseFlags(cmd *cobra.Command) error {
 	if err := viper.BindPFlag("nonce", cmd.PersistentFlags().Lookup("nonce")); err != nil {
 		return err
 	}
+	ConfigPath = viper.GetString("configPath")
+	if strings.Contains(ConfigPath, "../") {
+		return fmt.Errorf("😥 configPath should not contain traversal")
+	}
+	if stat, err := os.Stat(ConfigPath); !stat.IsDir() || os.IsNotExist(err) {
```
moshe-blox, line 314: This panics if there is stat is nil (may happen when there's an error afaik)

cli/initiator/initiator.go
```diff
@@ -30,76 +28,76 @@ var StartDKG = &cobra.Command{
 		██║  ██║██╔═██╗ ██║   ██║    ██║██║╚██╗██║██║   ██║   ██║██╔══██║   ██║   ██║   ██║██╔══██╗
 		██████╔╝██║  ██╗╚██████╔╝    ██║██║ ╚████║██║   ██║   ██║██║  ██║   ██║   ╚██████╔╝██║  ██║
 		╚═════╝ ╚═╝  ╚═╝ ╚═════╝     ╚═╝╚═╝  ╚═══╝╚═╝   ╚═╝   ╚═╝╚═╝  ╚═╝   ╚═╝    ╚═════╝ ╚═╝  ╚═╝`)
-		if err := cli_utils.SetViperConfig(cmd); err != nil {
-			return err
-		}
-		if err := cli_utils.BindInitFlags(cmd); err != nil {
-			return err
-		}
 		logger, err := cli_utils.SetGlobalLogger(cmd, "dkg-initiator")
```
moshe-blox, line 31: This fails because it tries to open LogFilePath flag but it has not yet been populated (its populated just below in BindInitFlag)

pkgs/crypto/crypto.go
y0sher: .

pkgs/initiator/initiator.go
```diff
@@ -102,13 +108,24 @@ type Payload struct {
...
 	// order the results by operatorID
 	sort.SliceStable(result, func(i, j int) bool {
 		return result[i].OperatorID < result[j].OperatorID
 	})
-
+	sorted := sort.SliceIsSorted(result, func(p, q int) bool {
+		return result[p].OperatorID < result[q].OperatorID
+	})
+	if !sorted {
+		return nil, fmt.Errorf("slice isnt sorted")
```
y0sher, line 127: not very informative error, which slice?

pkgs/initiator/initiator.go
```diff
@@ -153,15 +158,15 @@ func GeneratePayload(result []dkg.Result, sigOwnerNonce []byte) (*KeyShares, err
 		SharesData:  "0x" + hex.EncodeToString(sharesDataSigned),
 	}
 	ks := &KeyShares{}
-	ks.Version = "v4"
+	ks.Version = "v1.1.0"
```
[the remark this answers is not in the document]
y0sher: also maybe its best to set this in a constant/pass to the function for easier updating.

pkgs/initiator/initiator.go
```diff
@@ -197,6 +239,21 @@ func (c *Initiator) SendAndCollect(op Operator, method string, data []byte) ([]b
 	return resdata, nil
 }
 
+// SendAndCollect ssends http message to operator and read the response
+func (c *Initiator) GetAndCollect(op Operator, method string) ([]byte, error) {
```
y0sher, line 243: comment is outdated

pkgs/initiator/initiator.go
```diff
@@ -563,11 +567,30 @@ func (c *Initiator) StartDKG(id [24]byte, withdraw []byte, ids []uint64, network
...
+	}
+	resultMsg := &wire.ResultData{
+		Operators:          ops,
+		InitiatorPublicKey: pkBytes,
+		DepositData:        depositData,
+		KeysharesData:      keysharesData,
+	}
+	err = c.sendResult(consts.API_RESULTS_URL, id, ops, resultMsg)
+	if err != nil {
+		c.Logger.Error("🤖 Error storing results at operators", zap.Error(err))
```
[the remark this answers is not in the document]
y0sher: yeah I agree with @pavelkrolevets if ceremony was successful and its only the filesystem making trouble we shouldn't give up all the results from the ceremony.

pkgs/initiator/initiator.go
```diff
@@ -616,21 +676,36 @@ func (c *Initiator) StartReshare(newId, oldID [24]byte, oldIDs, newIDs []uint64,
...
 	c.Logger.Info("✅ verified master signature for ssv contract data")
+	// sending back to operators results
+	keysharesData, err := json.Marshal(keyshares)
+	if err != nil {
+		return nil, err
+	}
+	resultMsg := &wire.ResultData{
+		Operators:     newOps,
+		Identifier:    newId,
+		DepositData:   nil,
```
y0sher, line 702: so we're not sending the deposit data? only keysharesdata?

Makefile
```diff
@@ -49,6 +51,10 @@ docker-demo-reshare:
 	@echo "Running resharing in docker demo"
 	docker-compose up --build resharing
 
+docker-demo-ping:
+	@echo "Running resharing in docker demo"
```
y0sher, line 55: wrong text

pkgs/operator/state.go
```diff
@@ -29,9 +30,7 @@ import (
 const MaxInstances = 1024
 const MaxInstanceTime = 5 * time.Minute
 
-var ErrMissingInstance = errors.New("got message to instance that I don't have, send Init first")
-var ErrAlreadyExists = errors.New("got init msg for existing instance")
-var ErrMaxInstances = errors.New("max number of instances ongoing, please wait")
+type OperatorID *uint64
```
y0sher, line 33: never used, also why pointer to uint64?

pkgs/operator/operator.go
```diff
@@ -46,14 +48,18 @@ func (msg *KeySign) Decode(data []byte) error {
...
+const ErrTooManyHealthCheckRequests = `{"error": "too many requests to /health_check"}`
+const ErrTooManyResultRequests = `{"error": "too many requests to /results"}`
 
 // RegisterRoutes creates routes at operator to process messages incoming from initiator
 func RegisterRoutes(s *Server) {
 	// Add general rate limiter
 	s.Router.Use(httprate.Limit(
-		500,
-		1*time.Minute,
+		5000,
```
y0sher, line 61: We should probably have const

y0sher (approved):
only some minor comments

cli/utils/utils.go
pavelkrolevets: fixed

cli/initiator/initiator.go
pavelkrolevets: fixed

pkgs/initiator/initiator.go
pavelkrolevets: when resharing we get only a new `keyshares` file, validator pub and deposit stays the same, so no need to send it

pkgs/operator/state.go
pavelkrolevets: leftover, removed

pkgs/initiator/initiator.go
pavelkrolevets: fixed

pkgs/initiator/initiator.go
pavelkrolevets: fixed

Makefile
pavelkrolevets: fixed

pkgs/operator/operator.go
pavelkrolevets: fixed

pkgs/initiator/initiator.go
y0sher: oh I see this is only for resharing, missed that, thanks.

y0sher (approved):
Approving though build still not working.

moshe-blox:
One issue with the rate limit implementation: the initiator isn't aware when it hit that and will not retry.

We should probably not override the rate limit handler, because the default returns to the client `Retry-After` header which we can use in the http.Client to retry properly rather than fail.

pkgs/initiator/initiator.go
```diff
@@ -308,12 +367,9 @@ func (c *Initiator) MakeMultiple(id [24]byte, allmsgs [][]byte) (*wire.MultipleS
...
 func ValidatedOperatorData(ids []uint64, operators Operators) ([]*wire.Operator, error) {
-	if len(ids) < 4 {
-		return nil, fmt.Errorf("minimum supported amount of operators is 4")
-	}
-	// limit amount of operators
-	if len(ids) > 13 {
-		return nil, fmt.Errorf("maximum supported amount of operators is 13")
+	allowedLen := map[int]bool{4: true, 7: true, 10: true, 13: true}
+	if !allowedLen[len(ids)] {
+		return nil, fmt.Errorf("amount of operators should be 4,7,10,13")
```
y0sher, line 372: could've probably just check min/max 4/13) and that LEN%3 == 1 like we do in the contract

pavelkrolevets: I was thinking we are not using 10, so LEN%3 == 1 wouldnt be working.

pkgs/initiator/initiator.go
```diff
@@ -266,15 +264,37 @@ func (c *Initiator) GetAndCollect(op Operator, method string) ([]byte, error) {
 
 // SendToAll sends http messages to all operators. Makes sure that all responses are received
 func (c *Initiator) SendToAll(method string, msg []byte, operatorsIDs []*wire.Operator) ([][]byte, error) {
-	results, err := iter.Mapper[*wire.Operator, []byte]{MaxGoroutines: len(operatorsIDs)}.
-		MapErr(operatorsIDs, func(operator **wire.Operator) ([]byte, error) {
-			result, err := c.SendAndCollect(c.Operators[(*operator).ID], method, msg)
-			if err != nil {
-				return nil, fmt.Errorf("operator ID: %d, %w", (*operator).ID, err)
+	resc := make(chan opReqResult, len(operatorsIDs))
```
moshe-blox, line 267: Need to `defer close(resc)` here

PR #32, opened 2023-12-18
Resharing no db

pavelkrolevets:
Braking new resharing logic:

1. New flag `keysharesFilePath` to `keyshares` file
2. Get parameters needed for resharing from  this file 
4. Send reshare message with encrypted shares (`shares_data`) to all operators
5. Old participants decrypt the private share from `shares_data`  with ssv RSA pivate key used at initial ceremony
6. New participants recreate public commits from `shares_data` (from BLS pub keys).
7. Resharing DKG protocol begins
8. Result: new `keyshares` file stored at initiator and operators

Changes:
* removed `badger DB` from operators
* removed unnecessary flags from initiator to start resharing: `oldID`, `operatorIDs` , `owner`, `nonce`. We now get this values from `keyshares` file
* removed `instance_id` file storage at both inititator and operators as we dont need it anymore as a value at DB

TODO:
1. We need to identify initiator to be sure that the same initiator starts resharing. We removed DB where we stored his pub key. 
Suggested solutions: 
a) Add a field to keyshares file with operator sigs on initiator pub
b)  Store operator sigs on initiator pub at initiator


cli/utils/utils.go
```diff
@@ -382,26 +388,66 @@ func BindInitFlags(cmd *cobra.Command) error {
...
+		return err
+	}
+	if err := viper.BindPFlag("operatorsInfoPath", cmd.PersistentFlags().Lookup("operatorsInfoPath")); err != nil {
 		return err
 	}
 	if err := viper.BindPFlag("newOperatorIDs", cmd.PersistentFlags().Lookup("newOperatorIDs")); err != nil {
 		return err
 	}
-	if err := viper.BindPFlag("oldID", cmd.PersistentFlags().Lookup("oldID")); err != nil {
+	if err := viper.BindPFlag("keysharesFilePath", cmd.PersistentFlags().Lookup("keysharesFilePath")); err != nil {
```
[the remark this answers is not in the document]
pavelkrolevets: fixed

pkgs/initiator/initiator.go
```diff
@@ -115,6 +115,11 @@ type pongResult struct {
 	result []byte
 }
 
+type CeremonySigs struct {
+	OperatorIDs []uint64 `json:"operatorIds"` // operators IDs
+	Sigs        string   `json:"ceremonySigs"`
```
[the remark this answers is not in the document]
pavelkrolevets: we need to write it as json, imho hex string better

pkgs/dkg/drand.go
```diff
@@ -477,6 +480,19 @@ func (o *LocalOwner) postReshare(res *kyber_dkg.OptionResult) error {
...
+		return err
+	}
+	dataToSign := make([]byte, len(secretKeyBLS.Serialize())+len(encInitPub))
+	copy(dataToSign[:len(secretKeyBLS.Serialize())], secretKeyBLS.Serialize())
+	copy(dataToSign[len(secretKeyBLS.Serialize()):], encInitPub)
+	ceremonySig, err := o.signFunc(dataToSign)
+	if err != nil {
+		o.broadcastError(err)
+		return err
+	}
```
y0sher, line 495: as @MatusKysel mentioned earlier, this part repeats itself so I'd move it to a function to make sure it runs the same everywhere

pkgs/operator/operator_test.go
```diff
@@ -306,3 +315,165 @@ func TestWrongInitiatorSignature(t *testing.T) {
...
+		v := suite.G1().Point()
+		err = v.UnmarshalBinary(blsPub.Serialize())
+		require.NoError(t, err)
+		kyberPubhare := &share.PubShare{
+			I: int(i),
+			V: v,
+		}
+		kyberPubSharesFromPubs = append(kyberPubSharesFromPubs, kyberPubhare)
+	}
+	// require.Equal(t, kyberPubSharesFromPubs, kyberPubShares)
```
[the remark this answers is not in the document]
pavelkrolevets: checking, changed a bit.

pkgs/operator/state.go
```diff
@@ -670,3 +592,90 @@ func (s *Switch) MarshallAndSign(msg wire.SSZMarshaller, msgType wire.TransportT
...
+			return nil, fmt.Errorf("cant verify initiator public key")
+		}
+		// Find operator ID by PubKey
+		v := suite.G1().Scalar().SetBytes(secret.Serialize())
+		kyberPrivShare = &share.PriShare{
+			I: int(s.OperatorID),
+			V: v,
+		}
+	}
+	return kyberPrivShare, nil
```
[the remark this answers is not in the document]
pavelkrolevets: true, this func was for both old operators, and new operators where should be nil. Now logic changed, ill add error if not recovered.

cli/utils/utils.go
```diff
@@ -61,25 +59,38 @@ var (
...
 // SetViperConfig reads a yaml config file if provided
 func SetViperConfig(cmd *cobra.Command) error {
 	if err := viper.BindPFlag("configPath", cmd.PersistentFlags().Lookup("configPath")); err != nil {
 		return err
 	}
+	if err := viper.BindPFlag("configFileName", cmd.PersistentFlags().Lookup("configFileName")); err != nil {
+		return err
+	}
 	ConfigPath = viper.GetString("configPath")
+	configFileName := viper.GetString("configFileName")
```
y0sher, line 73: I meant this could be something simpler like that
```suggestion
	configFileName := viper.GetString("configFileName")
	if configFileName == "" {
	   configFileName = "config.yaml"
	   if cmd.Use != "start-operator" {
	       configFileName = fmt.Sprintf("%s.yaml", cmd.Use)
	   }
	}
	configPathYAML = fmt.Sprintf("%s/%s", ConfigPath, configFileName)
```

pkgs/initiator/initiator.go
```diff
@@ -608,115 +620,148 @@ func (c *Initiator) StartDKG(id [24]byte, withdraw []byte, ids []uint64, network
...
+		return nil, nil, err
+	}
 	reshare := &wire.Reshare{
 		OldOperators:       oldOps,
 		NewOperators:       newOps,
 		OldT:               uint64(oldThreshold),
 		NewT:               uint64(newThreshold),
-		OldID:              oldID,
 		Owner:              owner,
 		Nonce:              nonce,
```
y0sher, line 716: can the resharing nonce be the same as the previous nonce? doesn't make sense to me. a new nonce should probably be provided, otherwise we won't be able to use this to register the validator to SSV

y0sher (changes requested):
mostly the nonce bug, also lets please make sure this goes through thorough testing by @andrew-blox and @Roy-blox before merging.

cli/utils/utils.go
pavelkrolevets: Yeah, as @RaekwonIII  mentioned at https://github.com/bloxapp/ssv-dkg/pull/37 it turned out not a good idea to have fixed files to look for at `config` folder. I changed this approach back to explicitly providing path to a file `--configPath config.yaml`

pkgs/initiator/initiator.go
pavelkrolevets: True, good catch.

pavelkrolevets: Fixed. Added nonce to resharing as a CLI param.

examples/config/operator7.example.yaml
```diff
@@ -2,9 +2,9 @@ privKey: /data/operator7/encrypted_private_key.json
 privKeyPassword: /data/operator7/password
 port: 3030
 operatorID: 7
+storeShare: false
```
y0sher, line 5: so this PR means we don't store shares anymore right? so this option in the config should be deprecated I guess, in config files and the node code..

pkgs/crypto/crypto.go
```diff
@@ -587,3 +595,95 @@ func EncryptPrivateKey(priv []byte, keyStorePassword string) ([]byte, error) {
...
+			err := VerifyRSA(&opPrivateKey.PublicKey, dataToVerify, sig)
+			if err != nil {
+				continue
+			}
+			initVerify = true
+			break
+		}
+		if !initVerify {
+			return nil, fmt.Errorf("cant verify initiator public key")
+		}
```
y0sher, line 680: so you go through all the sigs but only one is enough to verify? shouldn't be opposite?

pkgs/utils/test_utils/test_utls.go
```diff
@@ -34,33 +29,19 @@ type TestOperator struct {
...
-		Path:      t.TempDir(),
-	})
-	require.NoError(t, err)
 	operatorPubKey := priv.Public().(*rsa.PublicKey)
 	pkBytes, err := crypto.EncodePublicKey(operatorPubKey)
-	require.NoError(t, err)
-	swtch := operator.NewSwitch(priv, logger, db, []byte(version), pkBytes, id)
+	if err != nil {
+		panic(err)
+	}
```
y0sher, line 43: more suitable to do this:
```suggestion
	require.NoError(t, err)
```

pkgs/utils/test_utils/test_utls.go
```diff
@@ -85,16 +66,13 @@ func CreateTestOperator(t *testing.T, id uint64, version string) *TestOperator {
...
-		Ctx:       context.Background(),
-		Path:      t.TempDir(),
-	})
 	require.NoError(t, err)
 	operatorPubKey := priv.Public().(*rsa.PublicKey)
 	pkBytes, err := crypto.EncodePublicKey(operatorPubKey)
-	require.NoError(t, err)
-	swtch := operator.NewSwitch(priv, logger, db, []byte(version), pkBytes, id)
+	if err != nil {
+		panic(err)
```
y0sher, line 73: more suitable to do this:
```suggestion
	require.NoError(t, err)
```

pkgs/utils/test_utils/test_utls.go
```diff
@@ -85,16 +66,13 @@ func CreateTestOperator(t *testing.T, id uint64, version string) *TestOperator {
...
+	}
```
y0sher, line 74: more suitable to do this:
```suggestion
	require.NoError(t, err)
```

examples/config/operator7.example.yaml
pavelkrolevets: yes, this is a leftover. Removed

pkgs/crypto/crypto.go
pavelkrolevets: Each operator verifies its own sig because we sign `operator`s secret + initiator pub` to have a relation between them . To verify for all I would need to hash it and send this hash in addition to sig, so more stuff to transfer

pkgs/crypto/crypto.go
```diff
@@ -587,3 +595,95 @@ func EncryptPrivateKey(priv []byte, keyStorePassword string) ([]byte, error) {
...
+		encInitPub, err := EncodePublicKey(initiatorPubKey)
+		if err != nil {
+			return nil, err
+		}
+		sigs := utils.SplitBytes(reshare.CeremonySigs, SignatureLength)
+		for _, sig := range sigs {
+			dataToVerify := make([]byte, len(secret.Serialize())+len(encInitPub))
+			copy(dataToVerify[:len(secret.Serialize())], secret.Serialize())
+			copy(dataToVerify[len(secret.Serialize()):], encInitPub)
```
y0sher, line 670: no need to serialize so many times
```suggestion
                       serialized := secret.Serialize()
			dataToVerify := make([]byte, len(serialized)+len(encInitPub))
			copy(dataToVerify[:len(serialized)], secret.Serialize())
			copy(dataToVerify[len(sserialized):], encInitPub)
```

pkgs/crypto/crypto.go
```diff
@@ -587,3 +595,95 @@ func EncryptPrivateKey(priv []byte, keyStorePassword string) ([]byte, error) {
...
+				continue
+			}
+			initVerify = true
+			break
+		}
+		if !initVerify {
+			return nil, fmt.Errorf("cant verify initiator public key")
+		}
+		// Find operator ID by PubKey
+		v := suite.G1().Scalar().SetBytes(secret.Serialize())
```
y0sher, line 682: see above comment
```suggestion
		v := suite.G1().Scalar().SetBytes(serialized)
```

pkgs/crypto/crypto.go
y0sher: it will probably be better to verify all operators in this ceremony, we might want to do it on another PR though. @moshe-blox , @MatusKysel wdyt?

moshe-blox: seems like the message we're verifying contains the share's BLS private key, so it would be impossible to verify other operator's signatures, though i'm not sure why that message couldn't contain the public key instead, care to explain? @pavelkrolevets 
*(i still have lots to catch up on with DKG)*

btw, is the order of signatures in the CeremonySigs slice deterministic (or provided elsewhere in the reshare message), so you can know at which index yours is? if so, we could verify just one.

regardless, we should move serializations/allocations out of the loop:
```go
dataToVerify := make([]byte, len(serialized)+len(encInitPub))
copy(dataToVerify[:len(serialized)], serialized)
copy(dataToVerify[len(serialized):], encInitPub)
for _, sig := range sigs {
	err := VerifyRSA(&opPrivateKey.PublicKey, dataToVerify, sig)
	if err != nil {
		continue
	}
	initVerify = true
	break
}
```

pkgs/crypto/crypto.go
```diff
@@ -587,3 +595,96 @@ func EncryptPrivateKey(priv []byte, keyStorePassword string) ([]byte, error) {
...
+		encInitPub, err := EncodePublicKey(initiatorPubKey)
+		if err != nil {
+			return nil, err
+		}
+		sigs := utils.SplitBytes(reshare.CeremonySigs, SignatureLength)
+		serialized := secret.Serialize()
+		for _, sig := range sigs {
+			dataToVerify := make([]byte, len(serialized)+len(encInitPub))
+			copy(dataToVerify[:len(serialized)], serialized)
+			copy(dataToVerify[len(serialized):], encInitPub)
```
y0sher, line 671: like moshe said , I don't see why this is inside the loop

pkgs/crypto/crypto.go
pavelkrolevets: @moshe-blox yeah true about serializations/allocations at loop, my bad. 
I was thinking about using public instead, but it needs to be redesigned. Ill create a ticket for it.

y0sher (approved):

moshe-blox (changes requested):
looks good, just a few comments

pkgs/crypto/crypto.go
```diff
@@ -587,3 +595,92 @@ func EncryptPrivateKey(priv []byte, keyStorePassword string) ([]byte, error) {
...
+	encryptedKeys := utils.SplitBytes(keyshares[pubKeysOffset:], len(keyshares[pubKeysOffset:])/len(oldOperators))
+	// try to decrypt private share
+	var kyberPrivShare *share.PriShare
+	// encrypted shares are ordered in increasing order
+	var position int
+	for i, op := range oldOperators {
+		if op.ID == operatorID {
+			position = i
+		}
+	}
```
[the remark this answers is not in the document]
pavelkrolevets: @moshe-blox yes, I didnt do the check here as at next line we will get an error that we cant decode if the position isnt found or wrong. But anyway ill add additional return error if we dont find it in the list.

pkgs/crypto/crypto.go
```diff
@@ -587,3 +595,92 @@ func EncryptPrivateKey(priv []byte, keyStorePassword string) ([]byte, error) {
...
+}
+
+func GetSecretShareFromSharesData(keyshares, initiatorPublicKey, ceremonySigs []byte, oldOperators []*wire.Operator, opPrivateKey *rsa.PrivateKey, operatorID uint64) (*share.PriShare, error) {
+	suite := kyber_bls12381.NewBLS12381Suite()
+	signatureOffset := phase0.SignatureLength
+	pubKeysOffset := phase0.PublicKeyLength*len(oldOperators) + signatureOffset
+	sharesExpectedLength := EncryptedKeyLength*len(oldOperators) + pubKeysOffset
+	if len(keyshares) != sharesExpectedLength {
+		return nil, fmt.Errorf("GetSecretShareFromSharesData: shares data len is not correct, expected %d, actual %d", sharesExpectedLength, len(keyshares))
+	}
```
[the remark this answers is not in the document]
pavelkrolevets: we need it only to inform users that something wrong with the  `keyshares` slice , security wise if something is wrong with keyshares, we will not be able to decrypt it anywhay, and even if one of `oldOperators` cant decrypt the whole protocol will fail and return error. Anyway Ill add more checks here to have it well informed.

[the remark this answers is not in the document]
moshe-blox: @pavelkrolevets oh i see, but yes "well-informed" is what i mean — its good to have descriptive errors so users can know they provided wrong order for example instead of 'invalid signature' which is hard to debug

pkgs/crypto/crypto.go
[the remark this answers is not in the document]
moshe-blox: @pavelkrolevets yes it would be to see error such as "could not find own operator ID" for debugging

pkgs/crypto/crypto.go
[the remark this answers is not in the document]
pavelkrolevets: @moshe-blox added additional checks for public key position, length etc.

pkgs/crypto/crypto.go
```diff
@@ -684,3 +665,48 @@ func GetSecretShareFromSharesData(keyshares, initiatorPublicKey, ceremonySigs []
...
+	for i, op := range oldOperators {
+		if operatorID == op.ID {
+			ids[operatorID] = i
+		}
+	}
+	// check if operator ID
+	position, ok := ids[operatorID]
+	if !ok {
+		return nil, 0, fmt.Errorf("GetSecretShareFromSharesData: operator not found among old operators: %d", operatorID)
+	}
```
moshe-blox, line 686:
```suggestion
	position := -1
	for i, op := range oldOperators {
		if operatorID == op.ID {
			position = i
			break
		}
	}
	// check if operator ID
	if position == -1 {
		return nil, 0, fmt.Errorf("GetSecretShareFromSharesData: operator not found among old operators: %d", operatorID)
	}
```

moshe-blox (approved):
LFTM, good job 👍

PR #37, opened 2024-01-09
Fix/improve cli

pavelkrolevets:
According to users complaints:
* add `configFileName` flag specify yaml file to look at `configPath` folder to have more flexibility
* create `outputPath` and `configPath` if not exist (default `./config` and `./output` folders)
* change default `./logs/debug.log` to `debug.log`
* fix log mistakes


cli/utils/utils.go
```diff
@@ -95,6 +91,25 @@ func SetViperConfig(cmd *cobra.Command) error {
...
+	case "start-operator":
+		// fileName is already set to "config.yaml" by default
+	}
+	if configFileName != "" {
+		configPathYAML = fmt.Sprintf("%s/%s", configPath, configFileName)
+	} else {
+		configPathYAML = fmt.Sprintf("%s/%s", configPath, fileName)
+	}
+	return configPathYAML
+}
```
y0sher, line 111: this code is kinda funny, I would for example check `configFileName` before doing the switch block, honestly I would check it out of the function and run this function only if its empty.

RaekwonIII:
I drafted an initial comment, but it sounded more convoluted and confusing than helpful. Let me try with command examples.


### Current operator command:
```sh
docker run --restart unless-stopped --name ssv_dkg -p 3030:3030 -v "/Users/massi/dev/ssv-dkg-operator/":/data -v "/Users/massi/dev/ssv-dkg-operator/output":/output -it "bloxstaking/ssv-dkg:latest" start-operator --configPath /data  --outputPath /output
```
This seems easy enough, but has the downside of the requirement that the config file is called `config.yaml` and it clashes with default name for ssv-node, for example.

### Previous operator command:
```sh
docker run --restart unless-stopped --name ssv_dkg -p 3030:3030 -v "/Users/massi/dev/ssv-dkg-operator/":/data -it "bloxstaking/ssv-dkg:latest" /app start-operator --configPath /data/operator.yaml
```
Because I could specify the configuration file as a path-and-name, I could have called it however I wanted, and then mapped the folder containing it. Or do as the ssv-node allows, which is map the file while renaming it (`-v /path/file-name.yaml:/data/config.yaml`)

### Current webapp initiator command:
```sh
docker run -v $(pwd):/data -it "bloxstaking/ssv-dkg:latest" init \ 
--owner ... \
--nonce ... \
--withdrawAddress ... \
--operatorIDs ... \
--operatorsInfo '...' \
--network holesky \
--generateInitiatorKeyIfNotExisting \
--configPath /data \
--logFilePath /data/debug.log \
--outputPath /data
```

I think `logFilePath` is out of place, and since I am providing all the necessary parameters as CLI flags, I don't think `--configPath` is necessary either, in my opinion.

### Previous webapp initiator command:
```sh
docker run -v $(pwd):/data -it "bloxstaking/ssv-dkg:latest" /app init \
--owner ... \
--nonce ... \
--withdrawAddress ... \
--operatorIDs ... \
--operatorsInfo '...' \
--network prater \
--generateInitiatorKey \
--outputPath /data
```
Not much to say, other than confirming the comment above.

pavelkrolevets:
Okay, looks like a `config` folder was not the best idea. Ill make it back to file path. 
Also, newly generated keys will be stored at output path.

pavelkrolevets:
Closing as merged to https://github.com/bloxapp/ssv-dkg/pull/32

PR #41, opened 2024-02-14
Rsa signature algo change

pavelkrolevets:
Today DKG is using PSS padding whereas SSV uses PKCS#1 v1.5

We should align DKG to use the same padding to keep it consistent (if they use the same private keys, they should probably use them in the same way).

Plus, the Authorized Endpoints PR uses PKCS#1 v1.5 currently so it wont even work for the current DKG before this switch.


pavelkrolevets:
closing it as its not the priority at the moment

PR #44, opened 2024-02-22
Additinal validations of keyshares file

pavelkrolevets:


pavelkrolevets:
why it was needed in the first place if we removed everything???

PR #47, opened 2024-03-01
enter resharing only once

MatusKysel:
We have a bug, where we can start re-sharing multiple times with the flag, we enter to resharing only once


pavelkrolevets:
just a comment, we dont need to worry `o.resharingStarted = false` after we finish resharing, as we always create a unique owner for every ceremony.

pavelkrolevets (approved):

PR #48, opened 2024-03-05
Fix drand error

pavelkrolevets:
- use at share indexes simple 0-12 values instead of IDs
- propagate failed ceremony error to initiator
- add same ops and chain reshare integration test, i.e. [1-4] ->[1-4] and chain: init [1-4]-> reshare [5-8]-> reshare [9-12]

### DIFF
--- a/integration_test/integration_test.go
+++ b/integration_test/integration_test.go
@@ -1,7 +1,6 @@
 package integration_test
 
 import (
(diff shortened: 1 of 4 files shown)


pavelkrolevets:
closing it as its not backward compatible. Fix is #49

PR #49, opened 2024-03-06
Fix drand error

pavelkrolevets:
Story: DKG wasnt able to finish resharing at stage servers, but worked perfectly locally + unit tests. Turned out that its wrongly assigned indexes at shares when we recover kyber share from `keyshares.json`. 

- fix share indexes when recover from `keyshare.json` (hint: ID-1). Should be same as we use when `init`
- propagate failed ceremony error to initiator. Now we get an error at inititator if DKG fails inside.
- add same ops and chain reshare integration test, i.e. [1-4] ->[1-4] and chain: init [1-4]-> reshare [5-8]-> reshare [9-12]

### DIFF
--- a/pkgs/crypto/crypto.go
+++ b/pkgs/crypto/crypto.go
@@ -620,7 +620,7 @@ func GetPubCommitsFromSharesData(reshare *wire.Reshare) ([]kyber.Point, error) {
 			return nil, err
 		}
 		kyberPubhare := &share.PubShare{
-			I: int(i),
+			I: int(reshare.OldOperators[i].ID - 1),
 			V: v,
 		}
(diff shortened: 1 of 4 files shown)


pkgs/crypto/crypto.go
```diff
@@ -620,7 +620,7 @@ func GetPubCommitsFromSharesData(reshare *wire.Reshare) ([]kyber.Point, error) {
 			return nil, err
 		}
 		kyberPubhare := &share.PubShare{
-			I: int(i),
+			I: int(reshare.OldOperators[i].ID - 1),
```
moshe-blox, line 623: how can we guarantee that the order of the pubkeys matches the order of the operators in the reshare message?

PR #73, opened 2024-03-19
Add tests

pavelkrolevets:
* Out of order operators (I.e 3,2,4,1) 
* Duplicates operators (I.e 1,2,3,3) 
* Operator ids larger than 1000 (related to sorting issue)


pkgs/crypto/keyshares.go
```diff
@@ -85,11 +85,11 @@ func ValidateKeysharesCLI(ks *wire.KeySharesCLI, operators []*wire.Operator, own
 func VerifyValidatorAtSharesData(ids []uint64, keyShares, expValPubKey []byte) error {
 	// Verify that IDs are unique and ordered.
 	if !sort.SliceIsSorted(ids, func(i, j int) bool { return ids[i] < ids[j] }) {
-		return fmt.Errorf("operators and not unique and ordered")
+		return fmt.Errorf("operators not unique or not ordered")
```
moshe-blox, line 88: this is only "not ordered"

pkgs/crypto/keyshares.go
```diff
@@ -85,11 +85,11 @@ func ValidateKeysharesCLI(ks *wire.KeySharesCLI, operators []*wire.Operator, own
...
 	}
 	for i := 0; i < len(ids)-1; i++ {
 		if ids[i] == ids[i+1] {
-			return fmt.Errorf("operators and not unique and ordered")
+			return fmt.Errorf("operators not unique or not ordered")
```
moshe-blox, line 92: this is only "not unique"

moshe-blox (approved):

pkgs/crypto/keyshares.go
pavelkrolevets: ids[i] < ids[j] -> strict less. Same IDs will result in error, but yes we check order then ill change error

pavelkrolevets:
accidentally pushed bulk test here, removed by force pushing. Bulk test will go in another PR. Closing and merging.

PR #80, opened 2024-03-20
Add bulk tests

pavelkrolevets:
* add bulk high level integration tests for 4/7/10/13 operators 1/10/100 validators ceremonies.


.github/workflows/go-tasks.yml
```diff
@@ -24,10 +24,10 @@ jobs:
         run: go build -v ./...
 
       - name: Test
-        run: go install gotest.tools/gotestsum@latest && gotestsum --format testname
+        run: go test ./... -timeout 99999s
```
pavelkrolevets, line 27:
```suggestion
        run: go test ./... -timeout 3600s
```

PR #89, opened 2024-03-28
release v2.1.0

MatusKysel:
- fixes
    - fix log error on exit
    - removed not audited clause
    - all audit issues 
- features
    - support of TLS/HTTPS
    - validation cmd `verify`


Dockerfile
```diff
@@ -18,6 +18,14 @@ COPY . .
 ENV CGO_ENABLED=1
 ENV GOOS=linux
 
+# Setup a directory for your certificates
+RUN mkdir /ssl
+
+# Generate a self-signed SSL certificate
+RUN openssl req -x509 -newkey rsa:4096 -sha256 -days 3650 -nodes \
+  -keyout /ssl/tls.key -out /ssl/tls.crt \
+  -subj "/C=CN/ST=GD/L=SZ/O=ssv, Inc./CN=*.ssv.com"
```
lior-blox, line 27: @MatusKysel do we have to add domain name? if so I would keep it as example.com or ssvlabs.io not ssv.com or anything else

README.md
```diff
@@ -388,7 +381,7 @@ Under the assumption that all the necessary files (`encrypted_private_key.json`,
 
 ```sh
 docker run --restart unless-stopped --name ssv_dkg -p 3030:3030  \
--v "<PATH_TO_FOLDER_WITH_CONFIG_FILES>":/data -it \
+-v "<PATH_TO_FOLDER_WITH_CONFIG_FILES>":/data -u `id -u $USER` -it \
```
lior-blox, line 384: @MatusKysel why do we need the user param?

examples/generate_certs.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+openssl genrsa -out ./initiator/rootCA.key 4096
+openssl req -x509 -new -key ./initiator/rootCA.key -subj "/C=CN/ST=GD/L=SZ/O=ssv, Inc./CN=*.ssv.com" -days 3650 -out ./initiator/rootCA.crt
+
+
+for i in $(seq 1 8);
+do
+    openssl genrsa -out ./operator${i}/operator${i}.key 2048 && \
+    openssl req -newkey rsa:2048 -nodes -keyout ./operator${i}/operator${i}.key -subj "/C=CN/ST=GD/L=SZ/O=ssv, Inc./CN=*.ssv.com" -out ./operator${i}/operator${i}.csr && \
```
lior-blox, line 10: @MatusKysel same thing here regarding domain name

pkgs/initiator/initiator_test.go
```diff
@@ -67,7 +73,7 @@ func TestStartDKG(t *testing.T) {
 	withdraw := common.HexToAddress("0x0000000000000000000000000000000000000009")
 	owner := common.HexToAddress("0x0000000000000000000000000000000000000007")
 	t.Run("happy flow", func(t *testing.T) {
-		intr, err := initiator.New(ops, logger, "v1.0.2")
+		intr, err := initiator.New(ops, logger, "v1.0.2", rootCert)
```
lior-blox, line 76: @MatusKysel not sure its a good idea to have hardcoded version here, could be "test.version" to avoid confusion

Dockerfile
moshe-blox: @MatusKysel @lior-blox 
we should understand what this means
is it a vulnerability if someone owns the domain we use here?

lior-blox: We don't own ssv.com, if we need to have a domain then ssvlabs.io is the correct one to use

pavelkrolevets (approved):

pkgs/initiator/initiator.go
```diff
@@ -96,8 +97,11 @@ func GenerateAggregatesKeyshares(keySharesArr []*wire.KeySharesCLI) (*wire.KeySh
 }
 
 // New creates a main initiator structure
-func New(operators wire.OperatorsCLI, logger *zap.Logger, ver string) (*Initiator, error) {
+func New(operators wire.OperatorsCLI, logger *zap.Logger, ver string, certs []string) (*Initiator, error) {
 	client := req.C()
+	// set CA certificates if any
+	client.SetRootCertsFromFile(certs...)
```
moshe-blox, line 103: even though it works with an empty slice, we should probably only call it `if len(certs) > 0`
otherwise a library upgrade might break it

Dockerfile
moshe-blox: @MatusKysel is there a way to not include any values at all? are these required? and if so, what are the defaults when you generate a self-signed certificate in the many tools that do so? (maybe we can copy them)

moshe-blox (approved):

pavelkrolevets (approved):

lior-blox (approved):

PR #100, opened 2024-04-23
Re-sign and re-share using proofs and EIP1271 signature

pavelkrolevets:
`Re-sign:`
* add resigning flow for creating new signatures over fields like `owner`, `nonce` using old BLS shares
* add flag `proofsFilePath: /data/initiator/output/ceremony-2024-04-22--16-36-19.354/proofs.json`
* if `proofs.json` consists of many proofs -> bulk resign with `nonce+1`
* add integration tests for resigning
* add docker demo tests for resigning

`Re-share:`
* add reshare with eip1271 signature verification according to spec
* add threshold of old operators support
* use types from dkg-spec and align
* add integration tests for resharing cases
* add docker demo for resharing
* update re-signing to spec (eip1271 sig)
* add more integration resign tests

### DIFF
--- a/pkgs/wire/types_json.go
+++ b/pkgs/wire/types_json.go
@@ -271,12 +271,6 @@ func (sd *ShareData) UnmarshalJSON(data []byte) error {
 	return nil
 }
 
-type ResingMessageJSON struct {
-	Operators []*Operator `json:"operators"`
-	Resign    *Resign     `json:"resign"`
-	Proofs    []*Proof    `json:"proofs"`
-}
-
 type Resign struct {
 	spec.Resign
 }
--- a/pkgs/crypto/keyshares.go
+++ b/pkgs/crypto/keyshares.go
@@ -147,11 +147,11 @@ func GetPubCommitsFromProofs(operators []*spec.Operator, proofs []*spec.SignedPr
 		if err != nil {
 			return nil, err
 		}
-		kyberPubhare := &share.PubShare{
+		kyberPubshare := &share.PubShare{
 			I: int(operators[i].ID - 1),
 			V: v,
 		}
-		kyberPubShares = append(kyberPubShares, kyberPubhare)
+		kyberPubShares = append(kyberPubShares, kyberPubshare)
 	}
 	pubPoly, err := share.RecoverPubPoly(suite.G1(), kyberPubShares, threshold, len(operators))
 	if err != nil {


pkgs/initiator/initiator.go
```diff
@@ -280,8 +293,103 @@ func (c *Initiator) StartDKG(id [24]byte, withdraw []byte, ids []uint64, network
...
+	ops, err := ValidatedOperatorData(ids, c.Operators)
+	if err != nil {
+		return nil, nil, nil, err
+	}
+	pkBytes, err := spec_crypto.EncodeRSAPublicKey(&c.PrivateKey.PublicKey)
+	if err != nil {
+		return nil, nil, nil, err
+	}
+	instanceIDField := zap.String("Ceremony ID", hex.EncodeToString(id[:]))
+	c.Logger.Info("🚀 Starting dkg ceremony", zap.String("initiator public key", string(pkBytes)), zap.Uint64s("operator IDs", ids), instanceIDField)
```
radiken, line 309: Should be "Starting resigning ceremony"

pkgs/initiator/initiator.go
```diff
@@ -280,8 +293,103 @@ func (c *Initiator) StartDKG(id [24]byte, withdraw []byte, ids []uint64, network
...
+		Identifier:    id,
+		DepositData:   depositData,
+		KeysharesData: keysharesData,
+		Proofs:        proofsData,
+	}
+	err = c.sendResult(resultMsg, ops, consts.API_RESULTS_URL, id)
+	if err != nil {
+		return nil, nil, nil, fmt.Errorf("🤖 Error storing results at operators %w", err)
+	}
+	return depositDataJson, keyshares, proofsArray, nil
```
radiken, line 382: The codes here after operators resigning are the same with **StartDKG** above (line 254-293), for better coding practice and future maintainance, maybe wrap it in another function like "processDKGResults"?

pkgs/operator/operator.go
```diff
@@ -189,3 +204,27 @@ func rateLimit(logger *zap.Logger, limit int) func(http.Handler) http.Handler {
...
+	if err != nil {
+		return nil, fmt.Errorf("operator %d, failed to read request body, err: %v", operatorID, err)
+	}
+	signedMsg := &wire.SignedTransport{}
+	if err := signedMsg.UnmarshalSSZ(rawdata); err != nil {
+		return nil, fmt.Errorf("operator %d, failed to unmarshal SSZ, err: %v", operatorID, err)
+	}
+	// Validate that incoming message has requested type
+	if signedMsg.Message.Type != reqMessageType {
+		return nil, fmt.Errorf("operator %d, received non-init message to init route, err: %v", operatorID, errors.New("not init message to init route"))
```
GalRogozinski, line 219: The error seems init specific...
but the from logic seems like it can be generic (not just init message)

pkgs/wire/types_json.go
```diff
@@ -263,6 +265,62 @@ func (sd *ShareData) UnmarshalJSON(data []byte) error {
 	return nil
 }
 
+type ResingMessageJSON struct {
+	Operators []*Operator `json:"operators"`
+	Resign    *Resign     `json:"resign"`
+	Proofs    []*Proof    `json:"proofs"`
+}
```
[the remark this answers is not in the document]
pavelkrolevets: Leftover, removed

pkgs/initiator/initiator.go
```diff
@@ -280,8 +293,103 @@ func (c *Initiator) StartDKG(id [24]byte, withdraw []byte, ids []uint64, network
...
+	if err != nil {
+		return nil, nil, nil, err
+	}
+	instanceIDField := zap.String("Ceremony ID", hex.EncodeToString(id[:]))
+	c.Logger.Info("🚀 Starting dkg ceremony", zap.String("initiator public key", string(pkBytes)), zap.Uint64s("operator IDs", ids), instanceIDField)
+	// make resign message
+	rMsg := &wire.ResignMessage{
+		Operators: ops,
+		Resign: &spec.Resign{ValidatorPubKey: proofs[0].Proof.ValidatorPubKey,
+			Fork:                  network.GenesisForkVersion(),
```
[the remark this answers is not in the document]
pavelkrolevets: PR was created before the fork. Now we need to incorporate fork.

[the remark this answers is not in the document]
pavelkrolevets: For now I think it can be genesis

```
// GenesisForkVersion returns the genesis fork version of the network.
func (n Network) GenesisForkVersion() phase0.Version {
	switch n {
	case PyrmontNetwork:
		return phase0.Version{0, 0, 32, 9}
	case PraterNetwork:
		return phase0.Version{0x00, 0x00, 0x10, 0x20}
	case HoleskyNetwork:
		return phase0.Version{0x01, 0x01, 0x70, 0x00}
	case MainNetwork:
		return phase0.Version{0, 0, 0, 0}
	default:
		logrus.WithField("network", n).Fatal("undefined network")
		return phase0.Version{}
	}
}
```

pkgs/dkg/drand.go
```diff
@@ -476,3 +477,108 @@ func (o *LocalOwner) GetCeremonySig(secretKeyBLS *bls.SecretKey) ([]byte, error)
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to generate deposit data with root %w", err)
+	}
+	// Sign.
+	depositPartialSignature := secretKeyBLS.SignByte(signingRoot[:])
+	if depositPartialSignature == nil {
+		return nil, fmt.Errorf("failed to sign deposit data with partial signature %w", err)
+	}
+	// Validate partial signature
+	if val := depositPartialSignature.VerifyByte(secretKeyBLS.GetPublicKey(), signingRoot[:]); !val {
```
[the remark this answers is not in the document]
pavelkrolevets: Some double checks after debugging are left here. We can remove it of course. We have tests covering BLS work/decoding/encoding. If you have more ideas please feel free to add.

pkgs/dkg/drand.go
```diff
@@ -476,3 +477,108 @@ func (o *LocalOwner) GetCeremonySig(secretKeyBLS *bls.SecretKey) ([]byte, error)
...
+		err = fmt.Errorf("partial deposit root signature is not valid %x", depositPartialSignature.Serialize())
+		return nil, err
+	}
+	// Encrypt BLS share for SSV contract
+	encryptedShare, err := o.encryptFunc([]byte(secretKeyBLS.SerializeToHexStr()))
```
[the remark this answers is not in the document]
pavelkrolevets: same here as above

pkgs/dkg/drand.go
```diff
@@ -476,3 +477,108 @@ func (o *LocalOwner) GetCeremonySig(secretKeyBLS *bls.SecretKey) ([]byte, error)
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to encrypt BLS share: %w", err)
+	}
+	// Sign SSV owner + nonce
+	data := []byte(fmt.Sprintf("%s:%d", eth_common.Address(r.Resign.Owner).String(), r.Resign.Nonce))
+	hash := eth_crypto.Keccak256([]byte(data))
+	sigOwnerNonce := secretKeyBLS.SignByte(hash)
+	// Verify partial SSV owner + nonce signature
+	val := sigOwnerNonce.VerifyByte(secretKeyBLS.GetPublicKey(), hash)
```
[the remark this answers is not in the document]
pavelkrolevets: same here as above answered

cli/initiator/resigning.go
```diff
@@ -0,0 +1,130 @@
...
+				// Create new DKG initiator
+				dkgInitiator, err := initiator.New(opMap.Clone(), logger, cmd.Version, cli_utils.ClientCACertPath)
+				if err != nil {
+					return nil, err
+				}
+				// Create a new ID
+				id := crypto.NewID()
+				nonce := cli_utils.Nonce + uint64(i)
+				// Perform the resigning ceremony
+				depositData, keyShares, proofs, err := dkgInitiator.StartResigning(id, operatorIDs, arrayOfSignedProofs[i], ethNetwork, cli_utils.WithdrawAddress.Bytes(), cli_utils.OwnerAddress, nonce)
```
moshe-blox, line 81: remind me if i'm wrong: wasn't the point of proofs to be able to verify the RSA signatures of operators as a confirmation that they approved that file to be used for resiging/resharing? @MatusKysel @pavelkrolevets

the spec has a `ValidateSignedProofs` func, maybe we can use it here right after loading the proofs?

moshe-blox (changes requested):

moshe-blox (changes requested):
well done, just a few comments/questions

i noticed the build CI fails, can you please check? @pavelkrolevets

pkgs/dkg/drand.go
```diff
@@ -203,23 +203,10 @@ func (o *LocalOwner) PostDKG(res *kyber_dkg.OptionResult) error {
...
-	}
 	// Sign SSV owner + nonce
 	data := []byte(fmt.Sprintf("%s:%d", eth_common.Address(o.data.init.Owner).String(), o.data.init.Nonce))
 	hash := eth_crypto.Keccak256([]byte(data))
 	sigOwnerNonce := secretKeyBLS.SignByte(hash)
-	// Verify partial SSV owner + nonce signature
-	val := sigOwnerNonce.VerifyByte(secretKeyBLS.GetPublicKey(), hash)
-	if !val {
-		return fmt.Errorf("partial owner + nonce signature isnt valid %x", sigOwnerNonce.Serialize())
-	}
```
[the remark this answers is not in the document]
pavelkrolevets: @GalRogozinski  suggested that its overkill to check sig after we created one, agree here, maybe its too much.

pavelkrolevets:
@moshe-blox @MatusKysel @GalRogozinski @radiken I merged `reshare` PR #102 to here

cli/initiator/resigning.go
pavelkrolevets: we have this check when we write result files to disk: https://github.com/ssvlabs/ssv-dkg/blob/re-sharing/pkgs/validator/validator.go#L135

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/bloxapp/ssv-dkg
 
-go 1.20
+go 1.21
```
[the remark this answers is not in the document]
pavelkrolevets: yes, we have a dep which was complaining on go >= 1.20, now seems its solved after the update. Ill leave only 1.23

pkgs/dkg/drand.go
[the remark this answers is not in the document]
moshe-blox: wdym after we've created one? we're taking this from file input no?

if so then it might have been created by someone else or manipulated since creation

cli/initiator/resigning.go
moshe-blox: yes, but this is a different check on the new proofs right?

i think we should also validate the proofs before resigning and resharing, so that we wont start a process that we know is gonna be invalid because the proofs are, and we can let the user know of this issue early

pkgs/utils/utils.go
```diff
@@ -103,3 +104,70 @@ func GetNonce(input []byte) []byte {
...
+	for _, op := range oldOperators {
+		if tmp[op.ID] != nil {
+			delete(tmp, op.ID)
+		}
+	}
+	for _, op := range tmp {
+		set = append(set, op)
+	}
+	return set
+}
```
moshe-blox, line 165: we're missing unit tests for these 3 funcs

these are important to test because they're quite hard to verify just by looking (imo)

moshe-blox (changes requested):

pkgs/utils/utils.go
pavelkrolevets: Okay agree. Ill make unit tests for these.

cli/initiator/resigning.go
pavelkrolevets: Ok, agree, added initial proofs validation at initiator before sending messages

pkgs/utils/utils.go
```diff
@@ -103,3 +105,111 @@ func GetNonce(input []byte) []byte {
...
+	if len < 4 {
+		return fmt.Errorf("wrong operators len: < 4")
+	}
+	if len > 13 {
+		return fmt.Errorf("wrong operators len: > 13")
+	}
+	if len%3 != 1 {
+		return fmt.Errorf("amount of operators should be 4,7,10,13: got %d", len)
+	}
+	return nil
```
moshe-blox, line 214:
```suggestion
	switch len {
	case 4,7,10,13:
		return nil
	default:
		return fmt.Errorf("amount of operators should be 4,7,10,13: got %d", len)
	}
```

moshe-blox (approved):

PR #117, opened 2024-09-04
fix amount marshaling and keyshare version

MatusKysel:
- removed dkg version from keyshare
- fixed issue with deposit data because of update of go-eth2-client
- fix keyshare encryption


pavelkrolevets (approved):

integration_test/reshare_eip1271_sig_test.go
```diff
@@ -153,7 +153,7 @@ func TestReshareInvalidContractSig(t *testing.T) {
 	sk, err := keystore.DecryptKey(jsonBytes, string(keyStorePassword))
 	require.NoError(t, err)
 	owner := eth_crypto.PubkeyToAddress(sk.PrivateKey.PublicKey)
-	signedProofs, err := wire.LoadProofs("./stubs/4/000001-0xb2c0176e14077e792d3d2f72001687564a572f05d65d25bd6edb8777a68bf6981aa1769c68455d701825dcf1cc3a0453/proofs.json")
+	signedProofs, err := wire.LoadProofs("./stubs/4/000001-0xaa57eab07f1a740672d0c106867d366c798d3b932d373c88cf047da1a3c16d0816ac58bab5a9d6f6f4b63a07608f8f39/proofs.json")
```
GalRogozinski, line 156: really, invalid changed as well?

pkgs/crypto/deposit_data.go
```diff
@@ -39,7 +39,7 @@ func BuildDepositDataCLI(network core.Network, depositData *phase0.DepositData,
 	depositDataJson := &wire.DepositDataCLI{
 		PubKey:                hex.EncodeToString(depositData.PublicKey[:]),
 		WithdrawalCredentials: hex.EncodeToString(depositData.WithdrawalCredentials),
-		Amount:                spec_crypto.MaxEffectiveBalanceInGwei,
+		Amount:                uint64(spec_crypto.MaxEffectiveBalanceInGwei),
```
GalRogozinski, line 42: silly go question...
is this casting needed?

I see `Gwei` is an alias for `uint64` and before everything compiled

examples/config/initiator.example.yaml
```diff
@@ -3,7 +3,7 @@ withdrawAddress: "0x81592c3de184a3e2c0dcb5a261bc107bfa91f494"
 owner: "0xDCc846fA10C7CfCE9e6Eb37e06eD93b666cFC5E9"
 nonce: 1
 network: "holesky"
-validators: 10
```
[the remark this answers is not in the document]
pavelkrolevets: This is example, imo does not matter 10 or 1. its probably easier to have demo with only 1 validator.

pkgs/crypto/deposit_data.go
pavelkrolevets: compiler is complaining `cannot use spec_crypto.MaxEffectiveBalanceInGwei (constant 32000000000 of type phase0.Gwei) as uint64 value in struct literalcompiler`

PR #121, opened 2024-09-16
Gnosis multisig

pavelkrolevets:
* Add Gnosis JS multisig examples
* Add correct reshare/resign JSON go marshalling to get same hash as at JavaScript json.strignify()
* Add unit tests to check validity


cli/flags/flags.go
```diff
@@ -34,6 +34,7 @@ const (
 	ethKeystorePath   = "ethKeystorePath"
 	ethKeystorePass   = "ethKeystorePass"
 	ethEndpointURL    = "ethEndpointURL"
+	signedReshare     = "signedReshare"
```
alan-ssvlabs, line 37: This is not used?

pavelkrolevets: not yet, I think we will create additional flags at bulk?

pavelkrolevets: imo we should have `bulkReshareJSON` and `bulkResignJSON` flags which can accept filepath or raw JSON with the structure like here https://github.com/ssvlabs/ssv-dkg/blob/gnosis_multisig/integration_test/stubs/reshare/bulk_reshare_msgs.json

PR #122, opened 2024-09-18
add load proofs from raw json

### LINKED ISSUE #119: Raw JSON support for proofs flag in reshare + resign
### Description
Reshare and resign should support proofs in a raw format in addition to the path option.

### Motivation
The path option is appropriate for users building a reshare/resign command for themselves. However, it is not a good solution for users coming from the web app. Web app users will upload their proofs.json in the UI, select new operators (in the case of reshare), sign the message, and expect a reshare/resign docker command that will be built using these inputs.

### Requirements
- [ ] Add a new proofs flag for raw JSON support
- [ ] Ensure the flag works with bulk operation

pavelkrolevets:
Closes #119 

* add flag `--proofsRawJSON` to load proofs directly from raw json string
* add tests


pavelkrolevets:
Strange. tests are passing locally. Maybe something wrong with CI?

GalRogozinski (approved):

pavelkrolevets:
closing it as `proofsString` flag was added with another commits.

Issue #114, opened 2024-09-04
Bulk Reshare


GalRogozinski:
## Description

Many validators reshare at once in parallel

## Motivation 

Some large stakers want to change clusters to all of their validators at once

## Problem

If different validators registered with a different nonce, then we can't use the contract's `bulkRegister` command.
Maybe we can take max-nonce?

## Requirements

- [ ] Change the cli to support bulk reshare
- [ ] Execute bulk reshare in parallel
- [ ] Write test cases

RaekwonIII:
Can you elaborate further on:
> If different validators registered with a different nonce, then we can't use the contract's `bulkRegister` command.
> Maybe we can take max-nonce?

The initial creation of keyshares (the `init` command, to start a ceremony) has a **starting nonce**, and correctly generates each consecutive share by increasing its nonce, then generates a single output file where all the shares are ordered by the nonce itself.

The `reshare` command should essentially do the same, in my opinion. The list of validator keys to be reshared can be counted, and a starting nonce is needed as an input, then, for all intents and purposes, the logic is the same as `init` , except for the necessity of decrypting the existing shares, and verifying provided proofs.

alan-ssvlabs:
Explaining here how bulk resign and reshare works now: 
A user signs a bulk reshare message (an array of reshare message), which is generated as a JSON file. There is a cli to support the user generating the file, by providing information including *proofs file path or stringnified proofs, old operators, new operators, owner, withdraw address, nonce, network*.

An example one with two reshare messages: 
```json
[...]

PR #123, opened 2024-09-19
Bulk reshare

alan-ssvlabs:
Solves issue #114 

Enable bulk reshare for users to run multiple reshare ceremonies for validators in the same cluster at once. To run bulk reshare, a user firstly generate a bulk reshare json file using the new cli command. Then the initiator signs the hash of the bulk reshare file and send it to the operators.


pkgs/initiator/initiator.go
```diff
@@ -529,6 +529,58 @@ func (c *Initiator) StartResharing(id [24]byte, oldOperatorIDs, newOperatorIDs [
...
+		zap.String("withdrawal", hex.EncodeToString(bulkReshareMsg.Reshares[0].Reshare.WithdrawalCredentials)),
+		zap.String("owner", hex.EncodeToString(bulkReshareMsg.Reshares[0].Reshare.Owner[:])),
+		zap.Uint64("nonce", bulkReshareMsg.Reshares[0].Reshare.Nonce))
+
+	allDepositData := make([]*wire.DepositDataCLI, 0)
+	allKeyShares := make([]*wire.KeySharesCLI, 0)
+	allSignedProofs := make([][]*wire.SignedProof, 0)
+	for i, reshare := range bulkReshareMsg.Reshares {
+		// Sign reshare message
+		sig, err := c.SignReshare(reshare, sk)
```
GalRogozinski, line 556: how this works with multisig?

pkgs/operator/handlers.go
```diff
@@ -142,6 +142,28 @@ func (s *Server) reshareHandler(writer http.ResponseWriter, request *http.Reques
...
+	}
+
+	// hash the message and store it
+	bulkReshare := &wire.BulkReshareMessage{}
+	err = bulkReshare.UnmarshalSSZ(bulkReshareMsg.Message.Data)
+	if err != nil {
+		utils.WriteErrorResponse(s.Logger, writer, err, http.StatusBadRequest)
+		return
+	}
+	hash := eth_crypto.Keccak256(bulkReshareMsg.Message.Data)
```
GalRogozinski, line 161: this hash is just for internal storage in map with no security aspect?
we can cosnider a more lightweight function like muhash if so

PR #124, opened 2024-09-19
Fix: incorrect offset error

pavelkrolevets:


cli/utils/utils.go
```diff
@@ -84,7 +85,7 @@ func SetViperConfig(cmd *cobra.Command) error {
 	ConfigPath = viper.GetString("configPath")
 	if ConfigPath != "" {
 		if strings.Contains(ConfigPath, "../") {
-			return fmt.Errorf("😥 configPath should not contain traversal")
```
GalRogozinski, line 87: I feel 😢 the emoji is gone...
That's ok though 😛

GalRogozinski (approved):

PR #129, opened 2024-09-26
compare proofs

MatusKysel:


pavelkrolevets:
encrypted shares will be different each time we encrypt.

```
func rsa.EncryptPKCS1v15(random io.Reader, pub *rsa.PublicKey, msg []byte) ([]byte, error)
EncryptPKCS1v15 encrypts the given message with RSA and the padding scheme from PKCS #1 v1.5. The message must be no longer than the length of the public modulus minus 11 bytes.

The random parameter is used as a source of entropy to ensure that encrypting the same message twice doesn't result in the same ciphertext. Most applications should use [crypto/rand.Reader](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html) as random. Note that the returned ciphertext does not depend deterministically on the bytes read from random, and may change between calls and/or between versions.
```

PR #130, opened 2024-09-27
Mitigate malicious initiator ddos

pavelkrolevets:
Problem:
In InitInstance, any INIT message with a valid signature causes an entry to be added to the s.Instances array. If that array contains 1024 entries not older than 5 minutes, no new DKG can be initiated. Due to Signature verification allows arbitrary public key, anyone with network access can create valid INIT messages, easily denying service to the legitimate initiator.

Solution:
1. increase instances to max 1024 * 100. This will increase hardware requirements but to not critical point as its quite low now.
2. decrease rate limit to init/resign/reshare routes to 100.
3. decrease time to live for instances to 1 minute, as this is more than enough to finish all phases of DKG ceremonies.

In this case, to ddos an attacker needs >= 1024 different IP addresses.

### DIFF
--- a/pkgs/operator/operator.go
+++ b/pkgs/operator/operator.go
@@ -17,9 +17,14 @@ import (
 
 // request limits
 const (
-	generalLimit = 5000
-	routeLimit   = 500
-	timePeriod   = time.Minute
+	generalLimit          = 5000
+	initRouteLimit        = 100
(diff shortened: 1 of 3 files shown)


pavelkrolevets:
integration tests are failing because of request limits reach. We need to update tests to run not more than 100 requests in a minute.

PR #143, opened 2024-10-14
fix generation command

MatusKysel:
Fix generate cmd, so it will return hash instead of JSON that will be used for signing

### DIFF
--- a/cli/utils/utils.go
+++ b/cli/utils/utils.go
@@ -1080,12 +1080,12 @@ func Sync(logger *zap.Logger) error {
 
 func checkIfOperatorHTTPS(ops []wire.OperatorCLI) error {
 	for _, op := range ops {
-		url, err := url.Parse(op.Addr)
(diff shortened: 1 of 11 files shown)


pavelkrolevets (approved):

cli/initiator/reshare.go
```diff
@@ -86,26 +87,27 @@ var GenerateReshareMsg = &cobra.Command{
 				ethNetwork,
 				cli_utils.WithdrawAddress[:],
 				cli_utils.OwnerAddress,
-				nonce, 
+				nonce,
 				cli_utils.Amount,
 				signedProofs[i],
 			)
 			if err != nil {
-				logger.Fatal("😥 Failed to construct reshare message: ", zap.Error(err))
```
alan-ssvlabs, line 94: Maybe keep this log as it was because this error is constructing the actual message

alan-ssvlabs (approved):

PR #145, opened 2024-10-16
Fix amount check at deposit validate lost after merges

pavelkrolevets:
* fix amount check pectra ready lost during merges https://github.com/ssvlabs/ssv-dkg/pull/127/files#diff-3b86fb7881073f8442330d0299604fee3701b99ffd02af55d47546f24cc5bf91L124-L125
* add test with 2048 ETH


pavelkrolevets:
@MatusKysel this test https://github.com/ssvlabs/ssv-dkg/blob/fix-amount-deposit-validate/integration_test/reshare_threshold_old_ops_test.go#L26 is failing at CI but I cant reproduce at my machine. Can you please check?

PR #168, opened 2025-01-14
feat: don't exchange msg with old inactive operators

MatusKysel:
This PR improves how reshare flow handles inactive operators. Main improvement is that after first step of sending `SendReshareMsg` to all operators, we will filter out inactive ones from flow (if they are above threshold) and in second step we are not sending them `SendExchangeMsgs`.

## Why?
- improves overall speed of reshare in case of inactive operators
- in rare cases fix issue with timeouts

### DIFF
--- a/pkgs/initiator/initiator.go
+++ b/pkgs/initiator/initiator.go
@@ -287,13 +287,16 @@ func (c *Initiator) ReshareMessageFlowHandling(id [24]byte, signedReshare *wire.
 		reqIDs = append(reqIDs, msgID)
 	}
 	c.Logger.Info("sending signed reshare message to all operators")
-	var errs map[uint64]error
-	exchangeMsgs, errs, err := c.SendReshareMsg(id, signedReshare, allOps)
+	var initErrs map[uint64]error
+	exchangeMsgs, initErrs, err := c.SendReshareMsg(id, signedReshare, allOps)
 	if err != nil {
 		return nil, err
 	}
+	if len(initErrs) > 0 {
+		allOps = filterOpsWithErrors(allOps, initErrs)
+	}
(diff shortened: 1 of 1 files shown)


alan-ssvlabs (approved):
LGTM

pkgs/initiator/initiator.go
```diff
@@ -287,13 +287,16 @@ func (c *Initiator) ReshareMessageFlowHandling(id [24]byte, signedReshare *wire.
 		reqIDs = append(reqIDs, msgID)
 	}
 	c.Logger.Info("sending signed reshare message to all operators")
-	var errs map[uint64]error
-	exchangeMsgs, errs, err := c.SendReshareMsg(id, signedReshare, allOps)
+	var initErrs map[uint64]error
+	exchangeMsgs, initErrs, err := c.SendReshareMsg(id, signedReshare, allOps)
```
RaekwonIII, line 291: Possibly dumb question, and not very significant issue, if I am correct: we are sending the signed reshare message here, shouldn't `initErrs` be called `reshareErrs` instead?

RaekwonIII: I see you call this "init phase" below, so I guess there's that. 🤔

RaekwonIII (approved):
The fix looks ok. Only one minor comment on variable name (can be easily ignored)

RaekwonIII:
Hold on @MatusKysel , I just thought of something: I guess it's ok to exclude offline old operators from the list of operators the message is sent to.
But why not exclude all "old" operators, except the ones that are also part of the new operator set?

Isn't this a potential security risk as well? Like, these are included in the ceremony to create new shares, while they shouldn't, at all.

Let me know what you think about this, please.

GalRogozinski (approved):

PR #231, opened 2026-04-02
Fix healthcheck pong signature verification (use known operator RSA key)

julienh-ssv:
Fix for:

F-ssv-dkg-021 P2 Triaged
Pong Message Self-Verification is Circular
ssv-dkg · Security · Assigned: dev4 · Effort: —

The health check verified the pong signature using a public key extracted from the pong itself. An attacker could generate any key pair, sign a pong, and include the public key — it would always verify.

Impact: operator identity can be spoofed during health checks, misleading the initiator about which operators are online.

Breaking change (CLI):
- `ping` now requires a trusted operator list via `--operatorsInfo` or `--operatorsInfoPath` to verify operator identity.
- `--ip` is optional; if omitted, `ping` pings all operators from the provided operators list.

### DIFF
--- a/integration_test/health_check_test.go
+++ b/integration_test/health_check_test.go
@@ -1,6 +1,8 @@
 package integration_test
 
 import (
+	"net/http"
+	"net/http/httptest"
 	"testing"
 
 	"github.com/ethereum/go-ethereum"
@@ -9,8 +11,10 @@ import (
 	"go.uber.org/zap/zapcore"
 	"go.uber.org/zap/zaptest/observer"
 
+	spec_crypto "github.com/ssvlabs/dkg-spec/crypto"
 	"github.com/ssvlabs/dkg-spec/testing/stubs"
 	"github.com/ssvlabs/ssv-dkg/pkgs/initiator"
+	"github.com/ssvlabs/ssv-dkg/pkgs/wire"
 )
 
 func TestHealthCheck(t *testing.T) {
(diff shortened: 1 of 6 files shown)


integration_test/health_check_test.go
```diff
@@ -42,11 +46,78 @@ func TestHealthCheck(t *testing.T) {
...
+			},
+		}, zap.New(core), testVersion, nil, true)
+		require.NoError(t, err)
+
+		err = dkgInitiator.Ping([]string{ts.URL})
+		require.NoError(t, err)
+		matches := logs.FilterLevelExact(zapcore.ErrorLevel).FilterMessageSnippet("operator not healthy")
+		require.GreaterOrEqual(t, matches.Len(), 1, "expected spoofed identity to be rejected")
+	})
 }
```
olegshmuelov: The `spoofed_identity_rejected` test is solid, it covers the core vulnerability. One more case worth adding: operator ID mismatch. An attacker (or MITM) could replay a valid pong from operator A when we pinged operator B, the signature would verify against a different key, but the explicit ID check at line 816 is what catches this. A test that constructs a pong signed with the correct key but with a mismatched `pong.ID` would prove that guard works.

olegshmuelov (approved):
LGTM!

ljuba-ssv (approved):

momosh-ssv (approved):

PR #218, opened 2026-01-07
ci: automate Docker Hub releases with ARM64 support

kchojn:

PR #232, opened 2026-04-06
ci: add Docker Hub release workflow with ARM64 support

olegshmuelov:
## Summary
- Add `.github/workflows/docker-release.yml` — builds and pushes multi-arch (amd64 + arm64) Docker images to Docker Hub on `v*` tag push
- Fix `Dockerfile` alpine base image SHA pin from amd64-only platform digest to manifest list digest, enabling ARM64 builds

## Context
Supersedes the Docker Hub workflow portion of #218. The remaining changes in #218 (Makefile, README, GitLab CI) can be evaluated separately.

## Finding
F-ssv-dkg-034: Docker Hub CI missing — no docker image published on tag

## Test
- Verified golang base image SHA is already a manifest list (multi-arch safe)
- Verified alpine SHA was amd64-only; replaced with manifest list digest
- Full validation requires a test tag push

---

**UPDATE:** Bumped Go 1.25.8 → 1.25.9 across `ci.yml`, `releases.yml`, `Dockerfile`, and `go.mod` to resolve 3 stdlib CVEs flagged by `govulncheck` (GO-2026-4947, GO-2026-4946, GO-2026-4870 — `crypto/x509` and `crypto/tls`).

### DIFF
--- a/.github/workflows/docker-release.yml
+++ b/.github/workflows/docker-release.yml
@@ -1,0 +1,46 @@
+name: Docker Release
+
+on:
+  push:
+    tags:
+      - "v[0-9]+.*"
+
+concurrency:
+  group: docker-release-${{ github.ref }}
+  cancel-in-progress: false
+
+permissions:
+  contents: read
+
+jobs:
+  docker:
+    name: Build and push multi-arch Docker image
+    runs-on: ubuntu-latest
+    timeout-minutes: 45
+    steps:
+      - uses: actions/checkout@v6.0.2
+
+      - uses: docker/setup-qemu-action@v4.0.0
+        with:
+          platforms: linux/arm64
+
+      - uses: docker/setup-buildx-action@v4.0.0
+
+      - uses: docker/login-action@v4.1.0
+        with:
+          username: ssvlabs
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Build and push
+        uses: docker/build-push-action@v7.0.0
+        with:
+          context: .
+          platforms: linux/amd64,linux/arm64
+          push: true
+          tags: |
+            ssvlabs/ssv-dkg:${{ github.ref_name }}
+            ssvlabs/ssv-dkg:latest
+          build-args: |
+            VERSION=${{ github.ref_name }}
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
--- a/.github/workflows/docker-release.yml
+++ b/.github/workflows/docker-release.yml
@@ -29,7 +29,7 @@ jobs:
       - uses: docker/login-action@v4.1.0
         with:
           username: ssvlabs
-          password: ${{ secrets.DOCKERHUB_TOKEN }}
+          password: ${{ secrets.SSV_DKG_DOCKERHUB_TOKEN }}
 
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -28,7 +28,7 @@ RUN --mount=type=cache,target=/root/.cache/go-build \
     ./cmd/ssv-dkg
 
 # Final stage
-FROM alpine:3.21@sha256:22e0ec13c0db6b3e1ba3280e831fc50ba7bffe58e81f31670a64b1afede247bc
+FROM alpine:3.21@sha256:c3f8e73fdb79deaebaa2037150150191b9dcbfba68b4a46d70103204c53f4709
```
nkryuchkov, line 31: Why did the hash change? The version didn't

olegshmuelov: The old SHA (`22e0ec...`) was a platform-specific digest (amd64 only). Multi-arch builds with `--platform linux/arm64` would pull the wrong image or fail. The new SHA (`c3f8e7...`) is the manifest list digest for the same alpine:3.21 — it lets Docker resolve the correct architecture automatically.

.github/workflows/docker-release.yml
```diff
@@ -0,0 +1,46 @@
...
+      - uses: docker/setup-qemu-action@v4.0.0
+        with:
+          platforms: linux/arm64
+
+      - uses: docker/setup-buildx-action@v4.0.0
+
+      - uses: docker/login-action@v4.1.0
+        with:
+          username: ssvlabs
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
```
vaclav-ssvlabs: please use `SSV_DKG_DOCKERHUB_TOKEN`

vaclav-ssvlabs (dismissed):
LGTM, token was added to the repo as `SSV_DKG_DOCKERHUB_TOKEN` - so please adjust.

Only small note - do we want to also push `ssvlabs/ssv-dkg:latest`? As you mentioned in one of PRs for `ssv-oracle`, this could be dangerous in case we need to do a patch for an older release, and this would override the `latest` with an older version. 


It is more of a question of workflow if this scenario can ever happen.

I assume the multi-platform CI is expected to work, as I didn't test this myself and assume it works.

nkryuchkov (dismissed):

olegshmuelov:
> LGTM, token was added to the repo as `SSV_DKG_DOCKERHUB_TOKEN` - so please adjust.

updated to `SSV_DKG_DOCKERHUB_TOKEN`.
> Only small note - do we want to also push `ssvlabs/ssv-dkg:latest`? As you mentioned in one of PRs for `ssv-oracle`, this could be dangerous in case we need to do a patch for an older release, and this would override the `latest` with an older version.
> 
> It is more of a question of workflow if this scenario can ever happen.
> 
> I assume the multi-platform CI is expected to work, as I didn't test this myself and assume it works.

Keeping `latest` for now, consistent with ssv-oracle. If we ever need to patch an older release we can revisit, but for a DKG tool the scenario is unlikely.

vaclav-ssvlabs (approved):

zktaiga (approved):

y0sher (approved):

PR #238, opened 2026-04-23
Load multisig RPC configuration from environment - F-ssv-dkg-009

julienharbulot:
## Ticket

`F-ssv-dkg-009`

This fixes a security finding in `ssv-dkg` where an Alchemy Sepolia API key was hardcoded in `integration_test/multisig_test.go` and the Gnosis multisig example scripts under `utils/gnosis_multisig_examples`.

## Changes

- Replaced the hardcoded Ethereum RPC URL in the multisig integration tests with `ETH_RPC_URL` and skip the RPC-dependent tests when it is not set.
- Replaced the hardcoded RPC URL in the Gnosis multisig example scripts with runtime environment variable loading.
- Updated the example READMEs to document the env-based secret flow.
- Expanded `.gitignore` to ignore `*.env` files.

## Validation

- `GOPROXY=https://proxy.golang.org,direct go test -timeout=3600s ./integration_test/... -run 'Test(ReshareBulkJSONPArsing|ResignBulkJSONPArsing|VerifyMultisigSignedOnChain2of3|VerifyMultisigSignedOnChain|VerifyMultisigSignedOffChain|VerifyEOASigned)$'`
- `GOPROXY=https://proxy.golang.org,direct go tool golangci-lint run -v ./integration_test/...`

Fixes - Fixes ssvlabs/ssv-node-board#623

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,7 +30,6 @@ dkg_output*/
 examples/*/output/*
 
 # Secrets
-.env
 *.env
 
 # Node.js (for utils/ JS helpers)
--- a/.gitignore
+++ b/.gitignore
@@ -31,6 +31,7 @@ examples/*/output/*
 
 # Secrets
 .env
+*.env
 
 # Node.js (for utils/ JS helpers)
 node_modules/
--- a/integration_test/multisig_test.go
+++ b/integration_test/multisig_test.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"path/filepath"
 	"strconv"
+	"strings"
 	"testing"
 
 	"github.com/ethereum/go-ethereum/accounts/keystore"
@@ -18,7 +19,23 @@ import (
 	"github.com/ssvlabs/ssv-dkg/pkgs/wire"
 )
 
-const EthRPC string = "https://eth-sepolia.g.alchemy.com/v2/YyqRIEgydRXKTTT-w_0jtKSAH6sfr8qz"
+const ethRPCEnvVar = "ETH_RPC_URL"
+
+// requireEthereumClient skips RPC-dependent tests when no endpoint is configured locally.
+func requireEthereumClient(t *testing.T) *ethclient.Client {
+	t.Helper()
+
+	ethRPC := strings.TrimSpace(os.Getenv(ethRPCEnvVar))
+	if ethRPC == "" {
+		t.Skipf("%s is not set", ethRPCEnvVar)
+	}
+
+	client, err := ethclient.Dial(ethRPC)
+	require.NoError(t, err)
+	t.Cleanup(client.Close)
+
+	return client
+}
 
 func TestReshareBulkJSONPArsing(t *testing.T) {
 	t.Parallel()
@@ -73,8 +90,7 @@ func TestVerifyMultisigSignedOnChain2of3(t *testing.T) {
 	t.Parallel()
 	t.Run("valid Gnosis 3/3 miltisig signatures", func(t *testing.T) {
 		gnosisAddress := common.HexToAddress("0x0205c708899bde67330456886a05Fe30De0A79b6")
-		ethBackend, err := ethclient.Dial(EthRPC)
-		require.NoError(t, err)
+		ethBackend := requireEthereumClient(t)
 
 		var finalMsg []byte
 		message := []byte("I am the owner of this Safe account")
@@ -102,8 +118,7 @@ func TestVerifyMultisigSignedOnChain(t *testing.T) {
 	t.Parallel()
 	t.Run("valid Gnosis 3/3 miltisig signatures", func(t *testing.T) {
 		gnosisAddress := common.HexToAddress("0x0205c708899bde67330456886a05Fe30De0A79b6")
-		ethBackend, err := ethclient.Dial(EthRPC)
-		require.NoError(t, err)
+		ethBackend := requireEthereumClient(t)
 
 		var finalMsg []byte
 		message := []byte("I am the owner of this Safe account")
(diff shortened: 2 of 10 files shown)


integration_test/multisig_test.go
```diff
@@ -18,7 +19,23 @@ import (
 	"github.com/ssvlabs/ssv-dkg/pkgs/wire"
 )
 
-const EthRPC string = "https://eth-sepolia.g.alchemy.com/v2/YyqRIEgydRXKTTT-w_0jtKSAH6sfr8qz"
```
nkryuchkov, line 21: It will stay in the git history. Can we make sure we have invalidated this URL?

.gitignore
```diff
@@ -31,6 +31,7 @@ examples/*/output/*
 
 # Secrets
 .env
+*.env
```
nkryuchkov, line 33: Can we remove the one above?

nkryuchkov (dismissed):

ljuba-ssv (dismissed):

integration_test/multisig_test.go
julienharbulot: Will talk to Calvix about it

.gitignore
julienharbulot: Addressed in d36265e by removing the redundant `.env` ignore entry and keeping `*.env` as the single rule.

ljuba-ssv (dismissed):

nkryuchkov (dismissed):

integration_test/multisig_test.go
```diff
@@ -18,7 +19,23 @@ import (
...
+const ethRPCEnvVar = "ETH_RPC_URL"
+
+// requireEthereumClient skips RPC-dependent tests when no endpoint is configured locally.
+func requireEthereumClient(t *testing.T) *ethclient.Client {
+	t.Helper()
+
+	ethRPC := strings.TrimSpace(os.Getenv(ethRPCEnvVar))
+	if ethRPC == "" {
+		t.Skipf("%s is not set", ethRPCEnvVar)
```
momosh-ssv, line 33: Should we consider wiring `ETH_RPC_URL` as a GitHub Actions secret on the `test-integration` job? Without it set, `TestVerifyMultisigSignedOnChain2of3`, `TestVerifyMultisigSignedOnChain`, `TestVerifyMultisigSignedOffChain`, and `TestVerifyEOASigned` now silently skip — they previously ran against the hardcoded RPC and are the only coverage for `spec_crypto.VerifySignedMessageByOwner`, the EIP-1271 ownership check used by resign/reshare. Silent skip leaves the gap invisible in the CI summary.

utils/gnosis_multisig_examples/validate/index.ts
```diff
@@ -30,8 +39,7 @@ interface Config {
 //   - Owner 3: public address from OWNER3_PRIVATE_KEY
 //   - SAFE_WALLET: public of safe wallet address with 2/3 threshold
 const config: Config = {
-  RPC_URL:
-    "https://eth-sepolia.g.alchemy.com/v2/YyqRIEgydRXKTTT-w_0jtKSAH6sfr8qz",
+  RPC_URL: requireEnv("ETH_RPC_URL"),
```
momosh-ssv, line 21: What do you think about dropping `OWNER1_PRIVATE_KEY` / `OWNER2_PRIVATE_KEY` / `OWNER3_PRIVATE_KEY` from the `Config` interface and `config` object here? `requireEnv` is added but only called for `RPC_URL`; the owner fields stay as `""` and are never read by the script. The other three `index.ts` files in this directory call `requireEnv` for all four, so the divergence makes the pattern hard to follow and a future developer adding signing logic would silently get empty keys.

utils/gnosis_multisig_examples/safe_off_chain/README.md
```diff
@@ -5,12 +5,18 @@
...
+$ export ETH_RPC_URL=https://eth-sepolia.g.alchemy.com/v2/<your-key>
+$ export OWNER1_PRIVATE_KEY=<owner-1-private-key>
+$ export OWNER2_PRIVATE_KEY=<owner-2-private-key>
+$ export OWNER3_PRIVATE_KEY=<owner-3-private-key>
 $ yarn install
 $ npx tsc
 $ node index.js
 ```
 
+You can also keep those values in a local `*.env` file and source it before running the script.
```
momosh-ssv, line 20: Might be worth clarifying the "source it" instruction — a plain `KEY=value` `.env` file is not directly sourceable in bash. Suggest `set -a; source local.env; set +a` (or `export $(grep -v '^#' local.env | xargs)`) so users don't run the script and wonder why the variables are still unset. Same applies to the `safe_on_chain/README.md` and `safe_on_chain_2/README.md` siblings.

julienharbulot: Addressed in 8e8445c. I added the `set -a; source local.env; set +a` flow to all three Safe example READMEs so plain `KEY=value` env files are exported before running the scripts.

utils/gnosis_multisig_examples/validate/index.ts
julienharbulot: Addressed in 8e8445c. I removed the unused owner private key fields from the validate example config and also dropped the now-unused imports and variables so the script only models the RPC and Safe address it actually uses.

integration_test/multisig_test.go
julienharbulot: Addressed in 8e8445c. The integration workflow now wires `ETH_RPC_URL` from repository secrets, and the RPC-backed multisig tests now fail visibly under `CI=true` if the secret is missing instead of silently skipping. Locally they still skip when no RPC URL is configured.

PR #244, opened 2026-08-17
chore(deps): build with Go 1.25.13 and refresh x/net, x/text

olegshmuelov:
Dependency and toolchain hygiene. No functional change — no Go source is touched.

## What

- Build with Go **1.25.13** in the three places that decide what actually ships — `Dockerfile`, `.github/workflows/ci.yml`, `.github/workflows/releases.yml`
- Refresh `golang.org/x/net` and `golang.org/x/text`
- Raise the integration-test timeout, which was flaking
- Update the README's build prerequisite to match

## Why

On the pinned Go 1.25.9, `govulncheck` reports 11 advisories reachable from the operator's TLS server and RPC client — among them `GO-2026-5856` (crypto/tls Encrypted Client Hello privacy leak) and `GO-2026-4918` (HTTP/2 infinite loop). Go 1.25.13 clears all of them.

CI has not run on `main` since 5 May, so this is not visible on the default branch today; any PR opened now fails the `security` job.

## Verification

| check | before | after |
|---|---|---|
| `govulncheck ./...` | 11 reachable | **No vulnerabilities found** |
| `go build` / `go vet` | — | pass |
| `./pkgs/...` suite | — | pass |
| Docker image | `go1.25.9` | `go1.25.13` |

Built the image and read the version out of the binary rather than inferring it from the Dockerfile.

`go.mod`'s language floor stays at 1.25.9 deliberately — raising it would impose a minimum on everyone consuming this as a module for no benefit here.


nkryuchkov (changes requested):
README.md still mentions 1.25.9:

https://github.com/ssvlabs/ssv-dkg/blob/chore/go-1.25.13-deps/README.md?plain=1#L203
https://github.com/ssvlabs/ssv-dkg/blob/chore/go-1.25.13-deps/README.md?plain=1#L207
https://github.com/ssvlabs/ssv-dkg/blob/chore/go-1.25.13-deps/README.md?plain=1#L436
https://github.com/ssvlabs/ssv-dkg/blob/chore/go-1.25.13-deps/README.md?plain=1#L440

y0sher (dismissed):

olegshmuelov:
@nkryuchkov Fixed in 565ebb9, all four.

`go.mod` still reads `go 1.25.9`, and I left that on purpose: it's the language floor, not a build pin. Raising it would impose a minimum on anything importing this as a module without changing what we actually build with. The three places that decide that are the Dockerfile and the two workflows.

y0sher (approved):

nkryuchkov (approved):

MatheusFranco99 (approved):

nkryuchkov:
> Raising it would impose a minimum on anything importing this as a module without changing what we actually build with

@olegshmuelov why is this an issue? If anything imports this as a module and builds with 1.25.9, then the vulnerabilities will stay. It's a patch version update, and the latest Go version is 1.26.6, so I don't think updating to 1.25.13 would cause any issues for users

olegshmuelov:
> > Raising it would impose a minimum on anything importing this as a module without changing what we actually build with
> 
> @olegshmuelov why is this an issue? If anything imports this as a module and builds with 1.25.9, then the vulnerabilities will stay. It's a patch version update, and the latest Go version is 1.26.6, so I don't think updating to 1.25.13 would cause any issues for users

You're right, verified it: byte-identical binary, `go.sum` untouched, `tidy` stable. The only effect is a clear `requires go >= 1.25.13` on `GOTOOLCHAIN=local`.
Not pushing it here though, it'd dismiss the three approvals.

PR #245, opened 2026-08-17
fix: bind resign/reshare validation to operator and owner identity

olegshmuelov:
Fixes two issues in the resign and reshare paths.

## 1. Ceremony proofs were validated against a caller-supplied key

`LocalOwner.Resign` selected the operator entry to validate the ceremony proof against from the operator list carried in the message, matching on ID. Since that list is supplied by the caller, the proof could be checked against a key the caller chose rather than this operator's own.

It now validates against this operator's own public key, and rejects the message if the operator is absent from the set.

## 2. Batch authorisation checked only the first message

For batched resign and reshare, the owner's EIP-1271 signature was verified against `Messages[0].Proofs[0].Proof.Owner`, after which every message in the batch was executed. A batch could therefore carry messages belonging to a different owner than the one that authorised it.

The authorised owner is now taken from the message being authorised, and every message in the batch is checked against it before any ceremony runs. Each message's own operator set is used rather than the first message's.

Ceremony failures are masked to a generic error, with a carve-out for version mismatch.

### DIFF
--- a/integration_test/resign_integration_test.go
+++ b/integration_test/resign_integration_test.go
@@ -79,6 +79,42 @@ func TestInitResignChangeOwnerRejected(t *testing.T) {
 		require.NoError(t, err)
 		rMsgs := []*wire.ResignMessage{rMsg}
 		_, _, _, err = executeResign(t, clnt, rMsgs, sk)
-		require.ErrorContains(t, err, "invalid owner address")
+		// the owner signature is checked against the resign message owner, so a message
+		// naming a different owner than the one that signed it stops at that check
+		require.ErrorContains(t, err, string(wire.InitiatorErrorCodeCeremonyFailed))
+	})
+}
+
+func TestInitResignProofOwnerMismatchRejected(t *testing.T) {
+	t.Parallel()
+	env := setupDynamicTest(t)
+	clnt, err := initiator.New(env.ops, env.logger, testVersion, rootCert, false)
+	require.NoError(t, err)
+	withdraw := newEthAddress(t)
+	sk, err := eth_crypto.GenerateKey()
+	require.NoError(t, err)
+	owner := eth_crypto.PubkeyToAddress(sk.PublicKey)
+	skOtherOwner, err := eth_crypto.GenerateKey()
+	require.NoError(t, err)
+	otherOwner := eth_crypto.PubkeyToAddress(skOtherOwner.PublicKey)
+	t.Run("4 operators proof owner mismatch rejected", func(t *testing.T) {
+		id := spec.NewID()
+		depositData, ks, proofs, err := clnt.StartDKG(id, eth1Creds(withdraw), []uint64{11, 22, 33, 44}, "holesky", owner, 0, uint64(spec_crypto.MIN_ACTIVATION_BALANCE))
+		require.NoError(t, err)
+		err = validator.ValidateResults([]*wire.DepositDataCLI{depositData}, ks, [][]*wire.SignedProof{proofs}, 1, owner, 0, withdraw)
+		require.NoError(t, err)
+		signedProofs := toSpecSignedProofs(proofs)
+		rMsg, err := clnt.ConstructResignMessage(
+			[]uint64{11, 22, 33, 44}, signedProofs[0].Proof.ValidatorPubKey, "mainnet",
+			eth1Creds(withdraw), owner, 10, uint64(spec_crypto.MIN_ACTIVATION_BALANCE), signedProofs,
+		)
+		require.NoError(t, err)
+		// the message keeps the owner that signs the batch, only the proofs name another
+		for _, proof := range rMsg.Proofs {
+			proof.Proof.Owner = otherOwner
+		}
+		rMsgs := []*wire.ResignMessage{rMsg}
+		_, _, _, err = executeResign(t, clnt, rMsgs, sk)
+		require.ErrorContains(t, err, string(wire.InitiatorErrorCodeCeremonyFailed))
 	})
 }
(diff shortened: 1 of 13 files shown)


y0sher (approved):

MatheusFranco99 (commented):
Reviewed against the fix list from the disclosure writeup. Everything we asked for is here and implemented correctly.

**Verified**

- **Fix 1** — `drand.go:607-612` builds `self` from `o.OperatorSecretKey.PublicKey` and validates against that. `spec.GetOperator` now has **zero call sites in the repo**, so the pattern is eradicated rather than patched at the one site.
- **Bounds** — `drand.go:591` length guard before any indexing, plus the `len(Messages) == 0` guard on the resign side that reshare already had (that was a live `Messages[0]` deref).
- **Fix 1b** — the authorised owner now comes from `Resign.Owner` / `Reshare.Owner` and is anchored by the EIP-1271 verification; `checkBatchMessageOwner` (`instances_manager.go:310`) checks every message and every proof against it before any ceremony runs; `allOps` is per-message. The decoy-`Messages[0]` scope smuggle is closed.
- **Fix 2** — `ceremonyResponseError` masks unconditionally on `/resign` and `/reshare`. I checked that `/init` and `/dkg` keeping `sanitizeCeremonyError` is still sound: the only RSA decryption of caller-supplied ciphertext is `drand.go:615` and `GetSecretShareFromProofs`, both on masked routes. The attacker-controlled version echo is gone.
- **`middleware.Recoverer`** installed first in the chain.

Pinning the regression tests to observed behaviour rather than error wording (6d9e127) is the right call for tests guarding a masking fix.

**Three follow-ups, none blocking**

1. **Resign has no uniqueness/ordering check on `r.Operators`, while reshare now has one** via `validateReshareStructure`. `ValidateResignMessage` in dkg-spec v1.2.0 never receives the operator list, so there is no such check anywhere on the resign path. Not exploitable after the `self` fix — the proof at `position` has to verify under our own key — but it is now the only structural check reshare has that resign doesn't, and that asymmetry between the two flows is what produced the original divergence. One `spec.UniqueAndOrderedOperators(r.Operators)` in `Resign` makes them symmetric.

2. **`state.go:92` — the reshare path's safety rests on a shadowed loop variable.** Inside `for i, op := range reshareMsg.Reshare.OldOperators`, `op` is shadowed by `op := &spec.Operator{ID: s.OperatorID, PubKey: s.PubKeyBytes}` before `ValidateReshareMessage`. That shadow is exactly why reshare never had this bug, but it reads like a redundant reassignment, and "cleaning it up" would reintroduce the bug on the reshare path. Renaming it `self`, matching the new code in `Resign`, is zero behaviour change and makes the invariant self-documenting.

3. **go.mod stays on dkg-spec v1.2.0**, so the fix is entirely caller-side. That's the right call for this release — no spec release blocking it — but `ValidateResignMessage` still accepts a caller-chosen `operator`. Worth a line in the description or the advisory noting the spec-side fix is tracked separately, so this isn't read as the spec being safe for third-party implementers.

**One question:** is the timing side channel (`drand.go:615` decrypt, then `:620` BLS parse) deferred by agreement? Masking equalises the error content but not the timing. Just want it recorded as a deliberate deferral rather than something expected in this PR.

**On the release:** since #244 moved the toolchain, the tagged binary won't match the ECR one regardless of the Go files being identical — so the smoke test off this branch is the check that actually matters. Agree with doing it here rather than after the merge.

olegshmuelov:
> **One question:** is the timing side channel (`drand.go:615` decrypt, then `:620` BLS parse) deferred by agreement? Masking equalises the error content but not the timing. Just want it recorded as a deliberate deferral rather than something expected in this PR.

Yes, deliberate. This release is limited to the code operators have already been running for two weeks.

MatheusFranco99:
@olegshmuelov What do you think about (1) and (2) in:

> 1. Resign has no uniqueness/ordering check on r.Operators, while reshare now has one via validateReshareStructure. ValidateResignMessage in dkg-spec v1.2.0 never receives the operator list, so there is no such check anywhere on the resign path. Not exploitable after the self fix — the proof at position has to verify under our own key — but it is now the only structural check reshare has that resign doesn't, and that asymmetry between the two flows is what produced the original divergence. One spec.UniqueAndOrderedOperators(r.Operators) in Resign makes them symmetric.
> 
> 2. state.go:92 — the reshare path's safety rests on a shadowed loop variable. Inside for i, op := range reshareMsg.Reshare.OldOperators, op is shadowed by op := &spec.Operator{ID: s.OperatorID, PubKey: s.PubKeyBytes} before ValidateReshareMessage. That shadow is exactly why reshare never had this bug, but it reads like a redundant reassignment, and "cleaning it up" would reintroduce the bug on the reshare path. Renaming it self, matching the new code in Resign, is zero behaviour change and makes the invariant self-documenting.

olegshmuelov:
> @olegshmuelov What do you think about (1) and (2) in:

(1) is in the current push: `drand.go:598` checks `spec.UniqueAndOrderedOperators(r.Operators)` in `Resign`, before the position lookup. You're right that `ValidateResignMessage` never receives the operator list, so the check lives in ssv-dkg rather than the spec.

(2) Agreed, and it's the more important of the two: that shadow is why reshare was never exploitable, and it reads like a redundant reassignment, so a cleanup deleting it would reintroduce the bug. It's behaviour-neutral, but I'd rather the diff stay identical to what operators have been running for two weeks, so it stays out of this release.

MatheusFranco99 (approved):

GalRogozinski (approved):

